Issue #31, opened 2014-09-24
[Feature request] Add new aggregation rule to diagnostic_aggregator


garaemon:
diagnostic_aggregator initialize analyzers from rosparam and it cannot add new analyzing rule after initialization. If it supports addition of rules, it would be very useful.

It means that if we want to add new rule, we need to run another aggregator and robot_monitor with remapping like `/diagnostics -> /diagnostics_perception`.

trainman419:
I don't think an add/delete API on a per-item basis is really appropriate for the way that this should be used; it gives the user too much freedom to accidentally remove important diagnostics.

The nominal use-case that I see for this API would be wanting to bring up additional diagnostics as part of a capability, and bring those diagnostics down again when that capability is no longer needed.

Perhaps something based on a model like `bond` would be more appropriate?

guillaumeautran:
Has there be any progress made on this feature request? 
I am (selfishly) volunteering myself to do the grunt work here if we could come up with a agreement on the API.

guillaumeautran:
Would specifying an `analyzers.yaml` file in the `package.xml` `<export/>` tag be a possible avenue? When the node comes up, the diagnostic aggregator could dynamically load the new YAML file and update its configuration. 
There would not be any unload capability though...

trainman419:
This has been implemented in #43

PR #41, opened 2015-09-08
diagnostic_aggregator: Add functionality for dynamically adding, deleting and listing analyzers

heuristicus:
See ros/common_msgs/pull/79 for the corresponding services used for functionality implemented here. This pull request implements the functionality mentioned in #31, but in a slightly different way since parameters must be loaded into a namespace before the service is called. Currently this is done in the `modify_diagnostics` script.

When adding new analyzers, the namespace to load from is specified in the service call, which a wrapper on the `init` function of the `AnalyzerGroup` uses to initialise a `NodeHandle`. The standard `init` function is then called with this modified handle which causes analyzers to be loaded from the specified namespace.

Deletion is done by recursively looking through the analyzers that are present and checking if their path is in the namespace which was sent in the service request. This means that a whole analyzer tree can be deleted by specifying the top level namespace.

Listing recursively aggregates all the paths into a vector which is then returned via the service response.

Any input on the changes would be much appreciated. In particular, I'm not sure if my changes in [`analyzer.h`](https://github.com/heuristicus/diagnostics/blob/hydro-devel/diagnostic_aggregator/include/diagnostic_aggregator/analyzer.h) make sense. The only other way I could think of implementing those recursive functions for use in [`analyzer_group.cpp`](https://github.com/heuristicus/diagnostics/blob/hydro-devel/diagnostic_aggregator/src/analyzer_group.cpp) without doing it this way was to check the class of the analyzer that was being looked at, and that seemed like a worse option.

I've not written any proper tests yet, but there are test launch files (those with the `test_script` prefix) which allow you to see the functionality at work.

`test_script.launch` will launch the `diagnostic_aggregator` with some default stuff loaded. You can then list those intial analyzers by running `test_script_list.launch`, or just `rosrun diagnostic_aggregator modify_diagnostics list`. Note that this won't show the `expected` analyzers, just those which exist. To add some analyzers to the aggregator, use `test_script_add.launch` and then list again - you should see that more analyzers have been added. You can then run the `test_script_del.launch` or `rosrun diagnostic_aggregator modify_diagnostics del /First` to delete one of the groups. The deletion is recursive, but isn't displayed by these tests since there isn't a test yaml which has a `AnalyzerGroup`. For the record, if you run things as I suggest, things should look something like the following:

    roslaunch diagnostic_aggregator test_script.launch

Then in another terminal:

    $ rosrun diagnostic_aggregator modify_diagnostics list
    /
    /My Path
    $ roslaunch diagnostic_aggregator test_script_add.launch
    $ rosrun diagnostic_aggregator modify_diagnostics list
    /
    /My Path
    /
    /First
    /Second
    /Third
    $ rosrun diagnostic_aggregator modify_diagnostics delete /First
    [INFO] [WallTime: 1441704184.793848] Successfully removed analyzers from diagnostics
    $ rosrun diagnostic_aggregator modify_diagnostics list
    /
    /My Path
    /
    /Second
    /Third

An additional `/` character is added to the list for each `AnalyzerGroup` added to the `Aggregator`.

There is also a thread on the drivers SIG [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/ros-sig-drivers/2lF1Wdi5ys4).

### DIFF
--- a/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.h
+++ b/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.h
@@ -48,6 +48,9 @@
 #include <diagnostic_msgs/DiagnosticArray.h>
 #include <diagnostic_msgs/DiagnosticStatus.h>
 #include <diagnostic_msgs/KeyValue.h>
+#include <diagnostic_msgs/AddDiagnostics.h>
+#include <diagnostic_msgs/RemoveDiagnostics.h>
+#include <diagnostic_msgs/ListDiagnostics.h>
 #include "XmlRpcValue.h"
 #include "diagnostic_aggregator/analyzer.h"
 #include "diagnostic_aggregator/analyzer_group.h"
@@ -125,6 +128,9 @@ public:
 
 private:
   ros::NodeHandle n_;
+  ros::ServiceServer add_srv_; /**< AddDiagnostics, /diagnostics_agg/add_diagnostics */
+  ros::ServiceServer rem_srv_; /**< RemoveDiagnostics, /diagnostics_agg/remove_diagnostics */
+  ros::ServiceServer list_srv_; /**< ListDiagnostics, /diagnostics_agg/list_diagnostics */
   ros::Subscriber diag_sub_; /**< DiagnosticArray, /diagnostics */
   ros::Publisher agg_pub_;  /**< DiagnosticArray, /diagnostics_agg */
   ros::Publisher toplevel_state_pub_;  /**< DiagnosticStatus, /diagnostics_toplevel_state */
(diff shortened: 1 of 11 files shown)


jack-oquin:
Hydro is no longer a suitable release target for significant new features.

But, this idea seems worth discussing on the ros-sig-drivers mailing list.

heuristicus:
Can you suggest which branch I should submit the request on? The `master` is currently about 3 years behind, it seems.

I've posted a thread on the mailing list as per your suggestion.

jack-oquin:
I am not a maintainer of this repository, but its branches need some major clean-up. It appears that ``hydro-devel`` and ``indigo-devel`` are currently identical and ``master`` is down-level. Also, for some reason, ``hydro-devel`` is the default branch, which makes no sense.

IMHO, we should delete most of those old, stale branches; merge the up-to-date commits into ``master``; and make that the default branch. I'd be willing to do that myself, if suitably authorized.

stonier:
+1 to all of jack's comments.

trainman419:
I've been following the recommended branch naming of having one branch for each version of ROS, so that it's clear what is released into each version of ROS.

the last few releases have been bug fixes and patches, so they've gone into Hydro and Indigo, hence why the hydro-devel and indigo-devel branches are the same.

I'll go through and do some branch cleanup tonight; then I'll have a branch to merge this into.

trainman419:
I've cleaned up most of the old branches, and updated master to be in sync with indigo-devel; feel free to resubmit this against master.

The changes here are substantial, but backwards-compatible; I think we should consider this for release into Indigo.

heuristicus:
New request at #42.

PR #43, opened 2015-10-02
diagnostic_aggregator: Add functionality for dynamically adding analyzers

heuristicus:
New proposal based on the bond framework as suggested by @trainman419.

The diagnostic aggregator now has a single service, which receives information about the namespace from which to load aggregator information. A new `AnalyzerGroup` is created and added to the main `AnalyzerGroup` of the aggregator. At the same time, a bond is formed between the node which requested the diagnostics be added and the aggregator. When this bond is broken, the additional diagnostic capabilities that were added are removed.

Simple tests can be run with the following commands:

```
roslaunch diagnostic_aggregator aggregator.launch
rosrun rqt_robot_monitor rqt_robot_monitor
rosrun diagnostic_aggregator add_analyzers diagnostics/diagnostic_aggregator/test/all_basic_analyzers.yaml
rosrun diagnostic_aggregator add_analyzers diagnostics/diagnostic_aggregator/test/multiple_match_analyzers.yaml __name:=testing
```

In different terminals, of course. The `add_analyzers` script adds analyzers based on the given parameters in the yaml files. In this specific case, the robot monitor will display the additional groups `First` and `Second` from the first run of `add_analyzers`, and `Header1` and `Header2` from the second run. The parameters from those files are loaded into the node namespace, which is passed to the aggregator via a service. The aggregator uses that namespace to create a new analyzer group. Cancelling execution in those terminals will remove the added groups from the monitor.

The required service message is in ros/common_msgs#80.

There is a discussion at [`ros-sig-drivers`](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/ros-sig-drivers/2lF1Wdi5ys4).

### DIFF
--- a/diagnostic_aggregator/CMakeLists.txt
+++ b/diagnostic_aggregator/CMakeLists.txt
@@ -3,16 +3,13 @@ cmake_minimum_required(VERSION 2.8.3)
 project(diagnostic_aggregator)
 
 # Load catkin and all dependencies required for this package
-find_package(catkin REQUIRED diagnostic_msgs pluginlib roscpp rospy rostest xmlrpcpp)
-
-catkin_package(DEPENDS diagnostic_msgs pluginlib roscpp rospy xmlrpcpp
+find_package(catkin REQUIRED diagnostic_msgs pluginlib roscpp rospy rostest xmlrpcpp bondcpp)
+catkin_package(DEPENDS diagnostic_msgs pluginlib roscpp rospy xmlrpcpp bondcpp bondpy
     INCLUDE_DIRS include
     LIBRARIES ${PROJECT_NAME})
 
-find_package(Boost REQUIRED COMPONENTS system regex)
-include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
-
-include_directories(include ${catkin_INCLUDE_DIRS} gtest-1.7.0/include)
+find_package(Boost REQUIRED COMPONENTS system)
+include_directories(include ${catkin_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS} gtest-1.7.0/include)
 
 add_library(${PROJECT_NAME}
   src/status_item.cpp
@@ -39,6 +36,7 @@ target_link_libraries(analyzer_loader diagnostic_aggregator)
 
 if(CATKIN_ENABLE_TESTING)
   add_rostest(test/launch/test_agg.launch)
+  add_rostest(test/launch/test_add_agg.launch)
 
   # Test Analyzer loader
   add_rostest(test/launch/test_loader.launch)
@@ -46,6 +44,10 @@ if(CATKIN_ENABLE_TESTING)
   add_rostest(test/launch/test_multiple_match.launch)
 endif()
 
+catkin_install_python(
+        PROGRAMS scripts/add_analyzers
+        DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
 install(DIRECTORY include/${PROJECT_NAME}/
         DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
 )
--- a/diagnostic_aggregator/scripts/add_analyzers
+++ b/diagnostic_aggregator/scripts/add_analyzers
@@ -1,0 +1,69 @@
+#!/usr/bin/env python
+
+NAME='add_analyzers'
+
+import sys
+import argparse
+from bondpy import bondpy
+from diagnostic_msgs.srv import AddDiagnostics
+import rosparam
+import rospy
+
+class AddAnalyzers:
+
+    def __init__(self, args):
+        rospy.on_shutdown(self.remove_group)
+        self.bond = None
+        self.add_analyzers(args)
+
+    def remove_group(self):
+        if self.bond:
+            self.bond.shutdown()
+
+    def add_analyzers(self, myargv):
+        usage = """
+        allows you to dynamically add a group to the diagnostic aggregator,
+        which is automatically removed when this node dies. Analyzer configurations
+        can either be provided directly, or assumed that they are already can be
+        found on the ros parameter server (where roslaunch may have loaded them).
+        """
+        parser = argparse.ArgumentParser(description=usage)
+        parser.add_argument('analyzer_yaml', nargs='?', default=None)
+        parser.add_argument('-t', '--timeout', type=float, dest='timeout', default=None, help='time in seconds to wait for the diagnostic_agg service to come up before timing out. Default waits indefinitely')
+        args = parser.parse_args(myargv[1:])
+
+        namespace = rospy.resolve_name(rospy.get_name())
+
+        if args.analyzer_yaml is None:
+            # nothing to do - it will assume parameters are already loaded on
+            # the parameter server (usually via roslaunch)
+            pass
+        else:
+            paramlist = rosparam.load_file(args.analyzer_yaml, default_namespace=namespace)
+            for params, ns in paramlist:
+                rosparam.upload_params(ns, params)
+
+        self.bond = bondpy.Bond("/diagnostics_agg/bond", namespace)
+        self.bond.start()
+
+        try:
+            rospy.wait_for_service('/diagnostics_agg/add_diagnostics', timeout=args.timeout)
+            add_diagnostics = rospy.ServiceProxy('/diagnostics_agg/add_diagnostics', AddDiagnostics)
+            resp = add_diagnostics(load_namespace=namespace)
+            if resp.success:
+                rospy.loginfo(NAME + ' successfully added analyzers to diagnostic aggregator')
+            else:
+                rospy.logerr(NAME + ' did not add any analyzers to diagnostic aggregator: ' + resp.message)
+                rospy.signal_shutdown('')
+        except rospy.service.ServiceException:
+            rospy.logerr(NAME + ' service returned failure - missing aggregator or failed init of analyzer group?')
+            rospy.signal_shutdown('')
+        except rospy.ROSException:
+            rospy.logerr(NAME + ' add timed out while waiting for diagnostics_agg service, or ROS shutdown')
+            rospy.signal_shutdown('')
+
+        rospy.spin()
+
+if __name__ == '__main__':
+    rospy.init_node(NAME)
+    AddAnalyzers(rospy.myargv())
--- a/diagnostic_aggregator/src/aggregator.cpp
+++ b/diagnostic_aggregator/src/aggregator.cpp
@@ -51,19 +51,18 @@ Aggregator::Aggregator() :
     base_path_ = "/" + base_path_;
 
   nh.param("pub_rate", pub_rate_, pub_rate_);
-    
+
   analyzer_group_ = new AnalyzerGroup();
-  
+
   if (!analyzer_group_->init(base_path_, nh))
   {
     ROS_ERROR("Analyzer group for diagnostic aggregator failed to initialize!");
   }
-  
 
   // Last analyzer handles remaining data
   other_analyzer_ = new OtherAnalyzer();
   other_analyzer_->init(base_path_); // This always returns true
-
+  add_srv_ = n_.advertiseService("/diagnostics_agg/add_diagnostics", &Aggregator::addDiagnostics, this);
   diag_sub_ = n_.subscribe("/diagnostics", 1000, &Aggregator::diagCallback, this);
   agg_pub_ = n_.advertise<diagnostic_msgs::DiagnosticArray>("/diagnostics_agg", 1);
   toplevel_state_pub_ = n_.advertise<diagnostic_msgs::DiagnosticStatus>("/diagnostics_toplevel_state", 1);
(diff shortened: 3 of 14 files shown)


diagnostic_aggregator/CMakeLists.txt
```diff
@@ -3,9 +3,9 @@ cmake_minimum_required(VERSION 2.8.3)
 project(diagnostic_aggregator)
 
 # Load catkin and all dependencies required for this package
-find_package(catkin REQUIRED diagnostic_msgs pluginlib roscpp rospy rostest xmlrpcpp)
-
-catkin_package(DEPENDS diagnostic_msgs pluginlib roscpp rospy xmlrpcpp
+find_package(catkin REQUIRED diagnostic_msgs pluginlib roscpp rospy rostest xmlrpcpp bondcpp bondpy)
+set(CMAKE_CXX_FLAGS "-g ${CMAKE_CXX_FLAGS}")
```
gavanderhoorn: Just noticing this, but I don't think you want to unconditionally add debug symbols here. This would override `CMAKE_BUILD_TYPE` derived flags.

diagnostic_aggregator/scripts/add_analyzers
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python
+
+NAME='add_analyzers'
+
+import sys
+import optparse
+from bondpy import bondpy
+from diagnostic_msgs.srv import AddDiagnostics
+import roslib; roslib.load_manifest('rosparam')
```
gavanderhoorn: This should not be necessary anymore in `catkin` Python pkgs (see [rospy tutorials - Setting Up Your PYTHONPATH](http://wiki.ros.org/rospy_tutorials/Tutorials/PythonPath)).

diagnostic_aggregator/CMakeLists.txt
heuristicus: Oops, I added that for tests but forgot to remove it.

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -107,13 +106,77 @@ void Aggregator::diagCallback(const diagnostic_msgs::DiagnosticArray::ConstPtr&
...
+    boost::function<void(void)>(boost::bind(&Aggregator::bondBroken, this)),
+    boost::function<void(void)>(boost::bind(&Aggregator::bondFormed, this))
+    );
+  req_bond->start();
+
+  boost::shared_ptr<Analyzer> group(new AnalyzerGroup());
+
+  bonds_.push_back(req_bond); // bond formed, keep track of it
+
+  if (group->init(base_path_, req.load_namespace))
```
trainman419: I don't think we need a new version of `Analyzer::init()` for this; it's just as easy to create a temporary NodeHandle to specify the namespace here:

```
if (group->init(base_path_, ros::NodeHandle(req.load_namespace)))
```

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -107,13 +106,77 @@ void Aggregator::diagCallback(const diagnostic_msgs::DiagnosticArray::ConstPtr&
...
+  }
+}
+
+void Aggregator::bondFormed(){}
+
+bool Aggregator::addDiagnostics(diagnostic_msgs::AddDiagnostics::Request &req,
+				diagnostic_msgs::AddDiagnostics::Response &res)
+{
+  // Wait for connections on this bond. Wait until it's formed to put it in the
+  // bond vector
```
trainman419: Blocking here while waiting for the bond to form will block all processing of other service calls and message callbacks, including processing of messages on the diagnostics topic.

This setup should probably be done without blocking the rest of the ROS event loop, and it should probably include a timeout as well, in case the calling node dies and doesn't clean up properly (assuming that isn't already contained in the bond API)

trainman419: It's probably better to do this setup as a callback in `bondFormed`, and avoid blocking here.

trainman419:
Aside from the code comments, you still haven't put forward a set of use cases or a larger justification for your changes. What is the greater purpose that you're trying to achieve here?

diagnostic_aggregator/src/aggregator.cpp
heuristicus: The comment is slightly out of date from some previous attempts at implementing the functionality. I wanted to ensure that the bond was successfully formed in the service call, so that the information could be passed back to the caller. Using the `waitUntilFormed` function in the service call doesn't work, since once inside it there is no way to receive messages on the bond topic unless there is another thread which is calling `ros::spin()` periodically, so I abandoned that approach.

Doing the initialisation in `bondFormed` means that the service call can only provide minimal information on return; probably just whether an attempt was made to add parameters from a namespace which was used before. If the `AnalyzerGroup` fails to initialise, the only way that information can be passed back to the caller is by breaking the bond, which I don't think is ideal.

stonier:
@heuristicus these are good comments - let's review together and update this PR early next week.

diagnostic_aggregator/test/launch/test_script_add.launch
```diff
@@ -0,0 +1,3 @@
+<launch>
+  <node pkg="diagnostic_aggregator" type="add_analyzers" name="add_analyzers" args="$(find diagnostic_aggregator)/test/multiple_match_analyzers.yaml /test/namespace"/>
```
stonier: Should result in an easier scrip, and also follows the aggregator style to load the parameters with rosparam rather than using an argument.

diagnostic_aggregator/scripts/add_analyzers
```diff
@@ -0,0 +1,74 @@
...
+        paramlist = rosparam.load_file(args[0])
+        namespace = args[1]
+        for params, ns in paramlist:
+            rosparam.upload_params(namespace + '/' + ns, params)
+
+        self.bond = bondpy.Bond("/diagnostics_agg_bond", namespace)
+        self.bond.start()
+
+        try:
+            rospy.wait_for_service('/diagnostics_agg/add_diagnostics', timeout=options.timeout)
```
stonier: Prefer

```
rospy.wait_for_service('add_diagnostics', timeout=options.timeout)
```

or 

```
rospy.wait_for_service('~add_diagnostics', timeout=options.timeout)
```

and let the user more conveniently remap instead of hardcoded absolute references.

stonier: Looks like diagnostics already uses absolute references everywhere (one diagnostics to rule them all?) which makes it resistant to namespacing anyway. No point fighting it in this pull request.

diagnostic_aggregator/src/aggregator.cpp
stonier: In short, it looks like we have two approaches:

1. report the result of adding analysers on the good faith that the bond will form
2. report success on the good faith that the analysers will be successfully added when the bond forms

Both have a chance of leaving the system in an imperfect state.

How about:

**Service Callback**

Set up a bond formed callback with partial binding (below).

Check the validity with `group->init` and return that result, but refrain from calling `analyzer_group->AddAnalyzer` and `added_analyzers_.insert()`. Store the group in an `analyzers_to_be_added_` map (or similar).

**Bond Formed Callback**

Add an argument to this function so you can pass request specific information to the callback (loading namespace, or even the validated group itself). Then when you set it as a callback for the bond, do *partial binding*. Finally, add it to the analysers.

stonier: You could also use partial binding on the bondBroken callback to avoid having to do an awkward iteration.

stonier:
@heuristicus Tested on the robots this morning, seems there is a kind of race condition between the 'Other' group and the newly added analyzer group.

diagnostics_race

When launching a separate system, if these nodes publish their diagnostics before the analyzer group gets added then they get added to the 'Other' group. Looks like it doesn't dynamically shift them from that group when the new analyzer gets added.

heuristicus:
I think the topic you're looking at is a slow publishing topic which is why it's possible to notice this. I'm not sure how much of a problem it is though. All messages are analysed when they are received by the aggregator - there's no memory of having analysed a status item with a specific name before, so things won't get stuck. 

One option is to change things so that the `Other` analyser only runs when there isn't a bond waiting to be formed. Another would be to make the analysers that are waiting for their bond to be formed accessible, and start processing incoming data before the bond actually forms. This might result in those status items disappearing if the bond fails to form, though.

heuristicus:
The problem with the `Other` group being populated instead of the analyzer is
caused by the initialisation of the vector part of `matched_` in
`analyzer_group::match`, line [185](https://github.com/ros/diagnostics/blob/afe104c4bb79acfa735247b6a1ab51ad1ee041ef/diagnostic_aggregator/src/analyzer_group.cpp#L185).

The problem is best explained with an example. Our initial setup for the
diagnostic aggregator has three `GenericAnalyzers` (it doesn't matter what type,
actually), none of which match diagnostics with the name `match_me`. We then
have a node which publishes a `DiagnosticStatus` with `name=match_me`. When this
node publishes the status for the first time, we attempt to match the
`StatusItem` created from it to one of the analyzers in the aggregator. We
haven't yet seen an item with this name, so a new `vector<bool>` of size 3 (the
number of analyzers in the aggregator) is created. The vector's indices are then
set based on whether or not the analyzer at the corresponding index matches the
name. In the example case, none of them match, so all of the indices are
`false`.

Now, we receive a request to add a new analyzer, which just so happens to look
for the name `match_me`. The request is good, and the bond is successfully
formed. The size of the `analyzers_` vector is now 4, as we added the
`analyzer_group` for the requested namespace. The next time we receive a message
from the node publishing diagnostics with `match_me`, we again try to match the
`StatusItem`. This time, we already know the name of the item, so we look at the
`vector<bool>` which was populated when the item was first seen. All of the
values in that vector are `false`, so the `match` function returns false, and we
defer to the `Other` group to analyze this item, which should actually be
analyzed by the new `analyzer_group`, if we had only bothered to check it.

This scenario can happen easily if the node publishing diagnostics has been
sitting there for a long time, and is a race between the new `analyzer_group`
being added and the first diagnostic message being received if the request is
made in a launch file which starts the node that publishes the diagnostics.
Sometimes the diagnostic ends up in the `Other` group, sometimes in the correct
place.

heuristicus:
The problem with the `Other` group being populated instead of the analyzer is
caused by the initialisation of the vector part of `matched_` in
`analyzer_group::match`, line [185](https://github.com/ros/diagnostics/blob/afe104c4bb79acfa735247b6a1ab51ad1ee041ef/diagnostic_aggregator/src/analyzer_group.cpp#L185).

The problem is best explained with an example. Our initial setup for the
diagnostic aggregator has three `GenericAnalyzers` (it doesn't matter what type,
actually), none of which match diagnostics with the name `match_me`. We then
have a node which publishes a `DiagnosticStatus` with `name=match_me`. When this
node publishes the status for the first time, we attempt to match the
`StatusItem` created from it to one of the analyzers in the aggregator. We
haven't yet seen an item with this name, so a new `vector<bool>` of size 3 (the
number of analyzers in the aggregator) is created. The vector's indices are then
set based on whether or not the analyzer at the corresponding index matches the
name. In the example case, none of them match, so all of the indices are
`false`.

Now, we receive a request to add a new analyzer, which just so happens to look
for the name `match_me`. The request is good, and the bond is successfully
formed. The size of the `analyzers_` vector is now 4, as we added the
`analyzer_group` for the requested namespace. The next time we receive a message
from the node publishing diagnostics with `match_me`, we again try to match the
`StatusItem`. This time, we already know the name of the item, so we look at the
`vector<bool>` which was populated when the item was first seen. All of the
values in that vector are `false`, so the `match` function returns false, and we
defer to the `Other` group to analyze this item, which should actually be
analyzed by the new `analyzer_group`, if we had only bothered to check it.

This scenario can happen easily if the node publishing diagnostics has been
sitting there for a long time, and is a race between the new `analyzer_group`
being added and the first diagnostic message being received if the request is
made in a launch file which starts the node that publishes the diagnostics.
Sometimes the diagnostic ends up in the `Other` group, sometimes in the correct
place.

heuristicus:
The fix in 66aa2a3 is the bare minimum for the new functionality.

heuristicus:
Previous commit didn't work when analyzers were removed and added multiple times.

stonier:
@trainman419 I've now put this on all of our gophers (various robots) and been testing it thoroughly over the last week.

It's finally working awesomely and is at the point where I'd recommend it going in :neckbeard:

Just for reference, this is in tandem with [common_msgs/pull/80](https://github.com/ros/common_msgs/pull/80).

diagnostic_aggregator/CMakeLists.txt
```diff
@@ -3,9 +3,8 @@ cmake_minimum_required(VERSION 2.8.3)
 project(diagnostic_aggregator)
 
 # Load catkin and all dependencies required for this package
-find_package(catkin REQUIRED diagnostic_msgs pluginlib roscpp rospy rostest xmlrpcpp)
-
-catkin_package(DEPENDS diagnostic_msgs pluginlib roscpp rospy xmlrpcpp
+find_package(catkin REQUIRED diagnostic_msgs pluginlib roscpp rospy rostest xmlrpcpp bondcpp bondpy)
+catkin_package(DEPENDS diagnostic_msgs pluginlib roscpp rospy xmlrpcpp bondcpp bondpy
```
trainman419: bondpy isn't using during the C++ compilation process or by downstream C++ compiles, so this doesn't need to find_package bondpy.

trainman419:
Looks reasonable.

I have one question about the locking that is introduced in the aggregator: it looks like `mutex_` is protecting `bonds_` and `analyzer_group_` from being modified concurrently from different threads, but this is only enforced in `bondBroken()`; none of the other functions that modify `bonds_` or `analyzer_group_` use the new mutex. This seems like an oversight; thoughts?

heuristicus:
When I added that mutex I wasn't sure if it would be required, so I didn't go further at that point. Now, I think enforcing the protection on the two vectors you mention is important, although concurrent modification is relatively unlikely. I don't think locking will affect performance much, if at all, since the things that are running aren't particularly computationally expensive.

jack-oquin:
Better safe than sorry. Concurrency bugs that seldom happen are the hardest to find and fix.

heuristicus:
Agreed. I think I caught every use of the shared vectors that might cause issues, but I'd appreciate any input on the critical sections I've set up.

jack-oquin:
What you are doing looks reasonable at a high level, but I don't know this component well enough to say whether it is sufficient.

Since adding new diagnostics is not expected to happen frequently, it should be safe and sufficient to hold a big mutex lock for the full duration of the new service calls and also for the /diagnostics topic callback. 

There is no need to optimize concurrency at any lower level.

stonier:
I'm also not 100% exactly sure around breaking up the analysing of individual items in that loop and would be..just a little nervous guaranteeing it without going deeper.

+1 to jack's comment. Being safe and clear is more important when there is little runtime cost attached.

heuristicus:
I've changed the scoped lock to be over the whole loop.

trainman419:
Looks good with the updated locking.

For completeness I'd like to see a few unit tests or ros tests that cover the things you've seen so far: mostly validating that newly added aggregators become active, and that new diagnostics get moved out of the 'Other' category and into the correct category.

It would also be useful to see some threading stress tests to check that there aren't any lurking concurrency issues; I may try to write those myself.

heuristicus:
Added a basic test for the things you suggested. The build currently fails because `common_msgs` hasn't yet pulled my changes to `diagnostic_msgs`.

trainman419:
Awesome! It's hard to get Travis to run integration tests across multiple packages; I'll run them tomorrow on my dev machine at work.

trainman419:
The tests for this pass on my machine; I think it's ready to merge.

trainman419:
I'd like to merge and release this into Indigo; it's only added functionality, so that qualifies as a patch release, and I think the added functionality justifies any remaining risk.

heuristicus:
There's still the outstanding PR at ros/common_msgs#80 which is required for this to work.

heuristicus:
While writing some tutorial stuff I thought of something which might cause some trouble. When adding via a launcher, the analyzer parameters are loaded into the node namespace of the `add_analyzers` script. If you try to launch another node with the same name, the first node will shut down, and its analyzers will be removed. However, the parameters loaded into the first node's namespace are not removed before it shuts down. This means that when the parameters for the new node are loaded, the ones from the old node will be loaded at the same time.

I'm sure if this is a big problem, but it breaks the principle of least surprise. My initial thoughts were to delete parameters inside the node's namespace once it shuts down, but the other node is brought up and parameters are loaded before the shutdown request is sent to the other node, so deleting them would also delete the newly loaded parameters.

Any thoughts on this? It's something which isn't likely to happen very often in practice, but will have strange results when it does.

heuristicus:
I've added a tutorial to the wiki [here](http://wiki.ros.org/diagnostic_aggregator/Tutorials/Adding%20Analyzers%20at%20Runtime).

stonier:
@heuristicus anonymous node names (see [here](http://wiki.ros.org/roslaunch/XML)?

trainman419:
I believe this is ready for merge. Can you squash it down into a single commit?

heuristicus:
Squashed it down. Sorry for the delay.

trainman419:
Thanks! I know this was a long process; thanks for sticking with it!

trainman419:
I will do a release of this once Tully releases a new version of common_msgs, so that I can make sure that I have the right version dependencies in place.

heuristicus:
Thanks for all your suggestions along the way.

trainman419:
This has been released into Indigo and Jade, and will be available after the next build farm sync to the public repository.

PR #46, opened 2016-02-23
Aggregator private fields where changed to protected, so as to be acc…

jvlahou:
…essed from derived classes


trainman419:
I haven't had much motivation to look at this because it isn't well described.

Changing the order of fields within a class breaks the ABI; if you want this to get released into Indigo please do not change the field ordering.

trainman419:
Closing due to inactivity.

PR #49, opened 2016-06-10
Start bond after add_diagnostics service is available

safrimus:
Right now, it is possible for `diagnostic_aggregator` to not add analyzers from a `add_analyzer` node in some cases even though all nodes are up and running.

Bond has a connection timeout duration and so if the `diag_agg` node launches `<connection_timeout>` secs after the `add_analyzer` node has been launched (or vice versa), the analyzers from `add_analyzers` are not added as the bond is considered broken.

If we start bond after the `/diagnostics_agg/add_diagnostics` service is available, we know for sure that the `diag_agg` node has been launched.

### DIFF
--- a/diagnostic_aggregator/scripts/add_analyzers
+++ b/diagnostic_aggregator/scripts/add_analyzers
@@ -44,10 +44,10 @@ class AddAnalyzers:
                 rosparam.upload_params(ns, params)
 
         self.bond = bondpy.Bond("/diagnostics_agg/bond", namespace)
-        self.bond.start()
 
         try:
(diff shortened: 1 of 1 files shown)


heuristicus:
Thanks for this. Something I overlooked during testing (no doubt there are other things as well).

mikepurvis:
@trainman419 We've bloomed this internally from a fork, as it's a critical fix for us, but we're interested in getting it merged so that we can be back on upstream for this repo.

trainman419:
Released into Indigo, Jade and Kinetic:
- https://github.com/ros/rosdistro/pull/11832
- https://github.com/ros/rosdistro/pull/11833
- https://github.com/ros/rosdistro/pull/11834

Issue #57, opened 2017-03-09
`rosdiagnostic echo` command line


guillaumeautran:
Viewing aggregated diagnostics can be done using  rqt_robot_monitor application. This is great but requires launching RQT in a Xwindows session. Instead, the `rosdiagnostic` command very quickly let you visualize the active diagnostic on any robot without requiring heavy duty Xwindow library.

mikepurvis:
Closed via #58.

PR #58, opened 2017-03-09
Created a command to print rosdiagnostics to the console.

guillaumeautran:
Works very much like rostopic echo but instead automatically connects to the aggregated diagnostics and output in a friendly format the content of the diagnostic report.

Issue: https://github.com/ros/diagnostics/issues/57

---
This change is [Reviewable](https://reviewable.io/reviews/ros/diagnostics/58)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@
 *.project
 *.cproject
 *.pydevproject
+*~ 
 
--- a/rosdiagnostic/CMakeLists.txt
+++ b/rosdiagnostic/CMakeLists.txt
@@ -1,0 +1,9 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(rosdiagnostic)
+
+find_package(catkin REQUIRED)
+
+catkin_package()
+
+catkin_python_setup()
+
--- a/rosdiagnostic/package.xml
+++ b/rosdiagnostic/package.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>rosdiagnostic</name>
+  <version>1.8.10</version>
+  <description>Command to print aggregated diagnostic contents to the command line</description>
+
+  <author email="gautran@clearpath.ai">Guillaume Autran</author>
+  <maintainer email="gautran@clearpath.ai">Guillaume Autran</maintainer>
+  <license>BSD</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
+  <exec_depend>rospy</exec_depend>
+  <exec_depend>diagnostic_msgs</exec_depend>
+
+</package>
--- a/rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
+++ b/rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
@@ -1,0 +1,187 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2017, Clearpath Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of Willow Garage, Inc. nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+import argparse
+import os
+import re
+import rosgraph
+import rospy
+import socket
+import sys
+
+from datetime import datetime
+from dateutil.tz import tzlocal
+from diagnostic_msgs.msg import DiagnosticArray, DiagnosticStatus
+
+
+class ROSTopicException(Exception):
+    """
+    Base exception class of related errors
+    """
+    pass
+
+
+class ROSTopicIOException(ROSTopicException):
+    """
+    Errors related to network I/O failures
+    """
+    pass
+
+
+def _check_master():
+    """
+    Make sure that master is available
+    :raises: :exc:`ROSTopicException` If unable to successfully communicate with master
+    """
+    try:
+        rosgraph.Master('/rostopic').getPid()
+    except socket.error:
+        raise ROSTopicIOException("Unable to communicate with master!")
+
+
+class RosDiagnostics(object):
+
+    DIAG_MSG_LVL_STR_NOCOLOR = {
+        DiagnosticStatus.OK:    ' OK  ',
+        DiagnosticStatus.WARN:  'WARN ',
+        DiagnosticStatus.ERROR: 'ERROR',
+        DiagnosticStatus.STALE: 'STALE',
+    }
+
+    DIAG_MSG_LVL_STR_COLOR = {
+        DiagnosticStatus.OK:    '\033[92m OK  \033[0m',
+        DiagnosticStatus.WARN:  '\033[93mWARN \033[0m',
+        DiagnosticStatus.ERROR: '\033[91mERROR\033[0m',
+        DiagnosticStatus.STALE: '\033[94mSTALE\033[0m',
+    }
+
+    def __init__(self, options):
+        self._lvl_strings_map = self.DIAG_MSG_LVL_STR_NOCOLOR if options.nocolor else self.DIAG_MSG_LVL_STR_COLOR
+        self._options = options
+        rospy.Subscriber(options.topic, DiagnosticArray, self._diag_callback, queue_size=1)
+
+    def _get_ns(self, name):
+        # Split namespace by slashes, return the base namespace.
+        return '/'.join(name.split('/')[:-1])
+
+    def _get_non_leaf_statuses(self, statuses):
+        return [self._get_ns(s.name) for s in statuses]
+
+    def _get_leaf_statuses(self, statuses):
+        parent_namespaces = self._get_non_leaf_statuses(statuses)
+        return [s for s in statuses if s.name not in parent_namespaces]
+
+    def _print_begin_banner(self, msg):
+        if not self._options.follow:
+            dt = datetime.fromtimestamp(msg.header.stamp.to_sec(), tzlocal())
+            print('=====================================================================')
+            print('Diagnostics generated on: {}'.format(dt))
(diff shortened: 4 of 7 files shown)


rosdiagnostic/package.xml
```diff
@@ -0,0 +1,16 @@
...
+  <name>rosdiagnostic</name>
+  <version>1.8.10</version>
+  <description>The rosdiagnostic package</description>
+
+  <author email="gautran@clearpath.ai">Guillaume Autran</author>
+  <maintainer email="gautran@clearpath.ai">Guillaume Autran</maintainer>
+  <license>BSD</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
+  <exec_depend>rospy</exec_depend>
```
mikepurvis, line 12: Depend on the messages you need.

Also drop the `<export>` element; not needed.

And fill in a proper description, please.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
```diff
@@ -0,0 +1,234 @@
...
+    Make sure that master is available
+    :raises: :exc:`ROSTopicException` If unable to successfully communicate with master
+    """
+    try:
+        rosgraph.Master('/rostopic').getPid()
+    except socket.error:
+        raise ROSTopicIOException("Unable to communicate with master!")
+
+
+class CprRosDiagnostics(object):
```
mikepurvis: Rename?

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
```diff
@@ -0,0 +1,234 @@
...
+    Base exception class of rostopic-related errors
+    """
+    pass
+
+
+class ROSTopicIOException(ROSTopicException):
+    """
+    rostopic errors related to network I/O failures
+    """
+    pass
```
mikepurvis: Change these exception names/descriptions?

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def _get_leaf_statuses(self, statuses):
+        parent_namespaces = self._get_non_leaf_statuses(statuses)
+        return [s for s in statuses if s.name not in parent_namespaces]
+
+    def _print_begin_banner(self, msg):
+        dt = datetime.fromtimestamp(msg.header.stamp.to_sec(), tzlocal())
+        print('=====================================================================')
+        print('Diagnostic generated on: {}'.format(dt))
+        print('---------------------------------------------------------------------')
```
mikepurvis: This is okay, but another possibility is using empy templates for this kind of thing.

mikepurvis:
This is an upstreaming of a tool we've been using internally at Clearpath for several months and has been really great. Example invocation:

```
$ cpr_rosdiagnostics 
=====================================================================
Diagnostic generated on: 2017-03-09 20:07:57.865806+00:00
---------------------------------------------------------------------
[ ERROR ] /Lidar/xxxxxxx_node: Hardware Status - Abnormal status: Sensor is initializing.
[ WARN  ] /Computer/Bundle Version - Updated bundle available.
[ WARN  ] /Computer/CPU Hogs - CPU hogs Warning: ['camera_nodelet_manager']
[ WARN  ] /Computer/Robot Environment - Robot environment variables not set
=====================================================================
```

Or, to include statuses which are okay:

```
$ cpr_rosdiagnostics  -l0
=====================================================================
Diagnostic generated on: 2017-03-09 20:09:07.863433+00:00
---------------------------------------------------------------------
[ ERROR ] /Computer/CPU Hogs - CPU hogs Error: ['camera_nodelet_manager']
[ ERROR ] /Lidar/xxxxxxx_node: Hardware Status - Abnormal status: Sensor is initializing.
[ WARN  ] /Computer/Bundle Version - Updated bundle available.
[ WARN  ] /Computer/Robot Environment - Robot environment variables not set
[  OK   ] /Computer/CPU Usage - CPU Usage OK
[  OK   ] /Computer/Disk IO - Disk IO OK
[  OK   ] /Computer/Load Average - Load Average OK
```

Big 👍 from me.

mikepurvis (changes requested):

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
guillaumeautran: Not sure to understand what you mean by an empty template. Can you please explain?

rosdiagnostic/package.xml
guillaumeautran: Done.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
guillaumeautran: Lightly reworded...

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
guillaumeautran: Renamed.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
mikepurvis: Empy is a templating scheme used in a bunch of core ROS stuff. Simple example:

https://github.com/mikepurvis/catkin_tools_python/blob/aeceac7c4bdeb5577e02c749b81a4c1a97e59bfc/catkin_tools_python/create_python_package_xmls.py#L21-L37

rosdiagnostic/.gitignore
```diff
@@ -0,0 +1 @@
+*~
```
mikepurvis: Drop this file— if you like, add this pattern to the one in the repo root.

rosdiagnostic/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+find_package(catkin REQUIRED COMPONENTS)
+
+catkin_package()
+
+catkin_python_setup()
+
+install(PROGRAMS
+  scripts/rosdiagnostic
+  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
```
mikepurvis: Newline please, and you can drop `COMPONENTS` from `find_package` above, since you haven't got any.

mikepurvis: Also, you should test this locally (`catkin_make install; find install`), but I don't think you need to separately install the script— that's already covered by the `scripts` arg in `setup.py`.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
```diff
@@ -0,0 +1,234 @@
...
+                try:
+                    root = ElementTree(None, package_manifest)
+                    version = root.findtext('version')
+                except Exception:
+                    pass
+        except rospkg.ResourceNotFound:
+            version = '(unknown)'
+
+    print("{} {}".format(prog, version))
+    return version
```
mikepurvis: I know you've just copied this over from [rosversion](https://github.com/ros-infrastructure/rospkg/blob/master/scripts/rosversion), but it's a lot of logic for what it does. IMO not worth it— none of the other standard rosxxxxx tools do this, so the expectation of a user would be that if they want to know what version this tool is, they would do `rosversion rosdiagnostic` same as they would do `rosversion rostopic`.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
```diff
@@ -0,0 +1,234 @@
...
+        if self._options.min_level <= diag.level:
+            print('[ {} ] {} - {}'.format(self._lvl_strings_map[diag.level], diag.name, diag.message))
+            if self._options.detail:
+                print('    hardware_id: {}'.format(diag.hardware_id))
+                for kv in diag.values:
+                    print('    - {}: {}'.format(kv.key, kv.value))
+
+
+def _rosdiagnostic_cmd_echo(argv):
+    parser = argparse.ArgumentParser(description='CPR ROS Diagnostics')
```
trainman419: I think the description here should be `ROS Diagnostics` or `ROS Diagnostic Viewer` now that this is going to be an upstream tool.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
```diff
@@ -0,0 +1,234 @@
...
+        elif command == 'version':
+            _rosdiagnostic_cmd_version('rosdiagnostic')
+        else:
+            _fullusage()
+    except socket.error:
+        sys.stderr.write("Network communication failed. Most likely failed to communicate with master.\n")
+        sys.exit(1)
+    except rosbag.ROSBagException as e:
+        sys.stderr.write("ERROR: unable to use bag file: %s\n" % str(e))
+        sys.exit(1)
```
trainman419: This catch for rosbag exceptions seems unnecessary; I don't see `rosbag` used anywhere else in this module, or any command-line arguments for specifying a bag file.

trainman419:
A few comments, but overall this looks like a great little debug and analysis tool.

rosdiagnostic/.gitignore
guillaumeautran: Done.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
guillaumeautran: Ok, will drop this whole part.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
guillaumeautran: I missed this. thanks for pointing it out. 
Changed to 'ROS Diagnostic Viewer'

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
guillaumeautran: Removed.

guillaumeautran:
Build passed and all comment have been addressed (please verify).

.gitignore
```diff
@@ -2,4 +2,5 @@
 *.project
 *.cproject
 *.pydevproject
+*~
```
guillaumeautran, line 5: Note, this is for Emacs backup files...

guillaumeautran:
@trainman419 When do you think this can be merged?

mikepurvis:
Someone will need to do a tag/release of this package for [ROS Lunar](http://repositories.ros.org/status_page/ros_lunar_default.html) in the coming weeks regardless, so hopefully it will be merged as part of that release process.

trainman419:
Sorry; I missed the notification that this had been updated. I have a very busy day today but I'll take a look at this over the weekend.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
```diff
@@ -0,0 +1,200 @@
...
+        argv = sys.argv
+    # filter out remapping arguments in case we are being invoked via roslaunch
+    argv = rospy.myargv(argv)
+
+    # process argv
+    if len(argv) == 1:
+        _fullusage()
+    try:
+        command = argv[1]
+        if command == 'echo':
```
trainman419: Now that there's only one command, this seems like unnecessary typing.

I think this would be more compact and easier to use if the functionality of `_fullusage` was integrated into the ArgumentParser instance and all of the logic for switching commands was removed.

trainman419:
I'll do a release for Lunar once this is merged and geometry is released, and since this is backwards-compatible, this will probably go into previous releases too.

rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py
guillaumeautran: The intention was to remain similar to the other rosXXX commands (rostopic, rosnode, etc)
I can simplify this part and make it different from the other commands, no problem.

mikepurvis: Having the verb structure is nice for future expandability. I guess the question is what would some other verbs on a tool like this be?

guillaumeautran: That was my reasoning, But at the same time, I cannot think of any other commands there. `list`, `info` don't seem applicable in this case. 
`rosdiagnostic pub <diagnostics to publish for testing>` might be of value to validate the diagnostic aggregator configuration. Yet again, not sure...

mikepurvis: Maybe there are flags of the current `echo` implementation that would be more logical as separate verbs, eg:

- `list` - List all diagnostics by name, with optional filtering, etc.
- `echo` (or `show`?) - Show key/values on just one report, by name.
- `watch` (or `notify`?) - Follow the diagnostics topic, but only echo to the console on level changes.

guillaumeautran: Not a bad idea. I would be a bit concerned about functionality duplication as there is very little differences between `list` + filter and `echo`. 
Anyway, after an offline chat, we will keep it simple now and can always expand the command later.

One change that I do want to make is to replace the `--namespace NAMESPACE` switch to a `--topic TOPIC` switch where one can specify the exact topic name the diagnostic can come from. That way, it will be possible to read diag from the `/diagnostics_agg` as well as `/diagnostics` topic. 

@trainman419 any other thoughts?

guillaumeautran:
I've updated this pull request with all the latest changes requested.

trainman419:
Looks good! Thanks.

PR #60, opened 2017-04-05
make rostest in CMakeLists optional (ros/rosdistro#3010)

bulwahn:


mikepurvis:
LGTM. A logical companion to this change would be bumping the package.xml files to version 2 and moving rostest to be a `test_depend`.

PR #61, opened 2017-04-10
Remove warning for missing queue size specification

familiarquark:
PR to remove the warning due to missing queue size. 
Added a default value of 10 as this is diagnostics and would help not to lose info. Please let me know if this would cause issues with backward compatibility in any way.


mikepurvis:
👍

trainman419:
Looks great! Thanks!

PR #63, opened 2017-04-11
feat(common_diagnostics): Added CPU percentage monitor

reinzor:
CPU monitor that outputs the average CPU percentage and a percentage per
CPU. The user can specify the warning CPU percentage. When one CPU exceeds
this percentage, the diagnostics status is set to WARN.


trainman419:
This looks very useful!

I'd like to see tests to go with this. At a minimum, a small node that verifies that uses the relevant parts of the psutil API to make sure that it returns sane data, and a test that mocks out psutil and validates that the warning thresholds work as desired.

Once this merges, we should also update the diagnostics_common_diagnostics wiki page to document the new node.

reinzor:
ping

reinzor:
ping

trainman419:
Sorry it has taken me a while to get back to this; I'm having a difficult time understanding how your tests work.

diagnostic_common_diagnostics/test/test_cpu_monitor.py
```diff
@@ -0,0 +1,102 @@
...
+
+PKG = 'diagnostics_common_diagnostics'
+NAME = 'test_cpu_monitor'
+if __name__ == '__main__':
+    parser = ArgumentParser()
+    parser.add_argument('--mock', action="store_true", default=False)
+    parser.add_argument('--percentage', default=50, type=int)
+    args = parser.parse_known_args()[0]
+
+    if args.mock:
```
trainman419: Instead of having a --mock option to this program and having it do completely different things (and then seeing it used as both the node under test and the test driver), it would be much more readable if this functionality was split into two separate ROS nodes

reinzor: I do agree, just updated the PR

trainman419: Looks good! Now that the PSUtilMock is in cpu_monitor_psutil_mock.py , it doesn't look like it's needed in this file any more. If that's the case, please remove it (and the --mock argument) from here.

reinzor: Forgot to remove this from the test file, should be removed now. Thanks for paying attention :+1:

diagnostic_common_diagnostics/CMakeLists.txt
```diff
@@ -10,6 +10,15 @@ catkin_package()
...
+if(CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
+
+  # Test CPU Monitor
+  add_rostest(test/launch/test_cpu_monitor.launch)
+  add_rostest(test/launch/test_cpu_monitor_level_ok.launch)
+  add_rostest(test/launch/test_cpu_monitor_level_warn.launch)
+endif()
+
 install(PROGRAMS
```
trainman419, line 22: Should add cpu_monitor.py to the install list here so that it gets included in debs.

reinzor: Added

PR #70, opened 2017-12-12
feat(common_diagnostics): Added GPU monitor

rsmallegoor:
Added a GPU monitor that provides the GPU and memory load percentages and the percentage of used memory. Warning thresholds for each value may be specified.

### DIFF
--- a/diagnostic_common_diagnostics/CMakeLists.txt
+++ b/diagnostic_common_diagnostics/CMakeLists.txt
@@ -21,6 +21,7 @@ endif()
 
 install(PROGRAMS
   src/diagnostic_common_diagnostics/cpu_monitor.py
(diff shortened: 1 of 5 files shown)


rsmallegoor:
Fix for the failing check: [PR](https://github.com/ros/rosdistro/pull/16579)

tfoote (changes requested):
Note that this dependency is pip only which means that this would not be releasable into debian packages anymore which is a requirement for a package at this low of a level.

This might make more sense as a standalone package available for people using nvidia GPUs to install from source instead of being in the default diagnostics repository since this is only useful for nvidia users as well.

rsmallegoor:
@tfoote Thank you for your advice. Now that I understand the limitations I've decided to create a standalone package. I will make it available as soon as it is finished.

PR #72, opened 2018-01-30
adapt to new xmlrpcpp header location

mikaelarguedas:
Without this `diagnostic_aggregator` fails to compile from source using `catkin build` or `catkin_make_isolated --merge`.

Note that this is applicable only to kinetic and later at the moment. A request for having the same header location for all distros has been submitted [here](https://github.com/ros/ros_comm/issues/1315)

### DIFF
--- a/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.h
+++ b/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.h
@@ -51,7 +51,7 @@
 #include <diagnostic_msgs/DiagnosticStatus.h>
 #include <diagnostic_msgs/KeyValue.h>
 #include <diagnostic_msgs/AddDiagnostics.h>
-#include "XmlRpcValue.h"
+#include "xmlrpcpp/XmlRpcValue.h"
 #include "diagnostic_aggregator/analyzer.h"
 #include "diagnostic_aggregator/analyzer_group.h"
 #include "diagnostic_aggregator/status_item.h"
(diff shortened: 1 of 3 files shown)


trainman419:
This breaks the indigo build and I'm not interested in maintaining a separate fork of this package just for experimental build tools.

This probably isn't the only package that is affected by this, and it seems like something that should be fixed in xmprpcpp. It looks like xmlrpcpp already has some extra cmake files that attempt to address this: https://github.com/ros/ros_comm/blob/lunar-devel/utilities/xmlrpcpp/cmake/xmlrpcpp-extras.cmake.develspace.in . It seems like it shouldn't be hard to make those build files work properly in `catkin build` and `catkin_make_isolated --merged`

gavanderhoorn:
@mikaelarguedas: wasn't https://github.com/ros/ros_comm/pull/1261 supposed to fix this?

I believe I ran into this issue in https://github.com/ros/ros_comm/pull/962, and @dirk-thomas fixed the cmake extras file for that.

mikaelarguedas:
It looks like the follow-up PR I was referring to just got released in Lunar that solves the compilation issue. It will be backported to kinetic so this should start building again soon. so no need to change the include statements here :+1: 

> I believe I ran into this issue in ros/ros_comm#962, and dirk-thomas fixed the cmake extras file for that.

Indeed the issue has been fixed upstream, it was just not released yet and I didn't look closely that the install from source instructions defaulted to getting the code from the release repos and not the source repo.

Thanks you both for your feedback

PR #75, opened 2018-02-06
Fixing travis build

LoyVanBeek:
This uses the travis-CI template from https://github.com/felixduvallet/ros-travis-integration

It runs OK but it finds that one test fails?


trainman419:
This seems to be making a lot of unnecessary changes, and that makes it difficult to review.

I'm going to try to figure out which of these changes are necessary and make only those changes, rather than rewriting the entire CI file.

trainman419:
Fixed in #76 with a two-line change.

PR #76, opened 2018-02-07
Install python-catkin-pkg to fix Travis

trainman419:
Travis CI fails because the catkin_pkg module is not installed. Install it to fix travis.

This also reverts the previous hardcoding of the build environment to a deprecated type. now that Trusty builds seem to be working correctly on Travis again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: required
 dist: trusty
 language: generic
-group: deprecated-2017Q3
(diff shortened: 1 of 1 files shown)


LoyVanBeek:
Interesting. I tried installing python-catkin-pkg and that didn't work. It is a dependency of ros-indigo-catkin, so I'm not sure why this change fixes it.

PR #84, opened 2018-07-11
Make FrequencyStatus' name configurable

NilsHaukeBussas-TomTom:
I think this is pretty self-explanatory. In some cases you'd like to give the frequency status some other name. This ought to be supported.

### DIFF
--- a/diagnostic_updater/include/diagnostic_updater/update_functions.h
+++ b/diagnostic_updater/include/diagnostic_updater/update_functions.h
@@ -124,6 +124,17 @@ namespace diagnostic_updater
     }
 
       /**
+       * \brief Constructs a FrequencyStatus class with the given parameters.
+       */
+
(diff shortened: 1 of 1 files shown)


mikepurvis:
Yes please! :+1:

trainman419:
Unfortunately, adding an argument to a method is an ABI-breaking change and would require that this only go into future ROS releases.

If you restructure this as a new constructor that takes two arguments, it won't break ABI and will be releasable for current versions of ROS.

NilsHaukeBussas-TomTom:
I don't want to pressure you, just want to make sure you notice my changes: I updated the patch by amending the commit because it's not additional work but replaces the old implementation. I've been told this might not have triggered a message to you. So, if it didn't, here's the reminder.

trainman419:
Looks good! Thanks for the notification; github isn't good about sending notifications on code changes.

PR #85, opened 2018-08-01
Disable param polling

mgrrx:
We noticed that we get frequent getParam calls against the parameter server for I'd say no reason.
The comment says that the C++ version uses getParamCached but when you look at the code you can see that it was removed.
I don't see a use case where somebody would want to change the frequency during runtime anyway - so why not remove it?

### DIFF
--- a/diagnostic_updater/src/diagnostic_updater/_diagnostic_updater.py
+++ b/diagnostic_updater/src/diagnostic_updater/_diagnostic_updater.py
@@ -230,7 +230,7 @@ class Updater(DiagnosticTaskVector):
 
         self.last_time = rospy.Time.now()
         self.last_time_period_checked = self.last_time
-        self.period = 1
+        self.period = rospy.get_param("~diagnostic_period", 1)
 
         self.verbose = False
         self.hwid = ""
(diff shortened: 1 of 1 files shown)


trainman419:
It looks like the C++ API is still using getParamCached to retrieve the diagnostic period: https://github.com/ros/diagnostics/blob/indigo-devel/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.h#L516-L521 .

Do you see something else that makes you think this is unused?

mgrrx:
Sorry for coming back to this so late but we are using this in our forks for a long time now.

You are right, the C++ version uses getParamCached. Anyway, I think the current python implementation inside diagnostics is not efficient as it is polling like crazy. And since we still don't have https://github.com/ros/ros_comm/pull/1515 for reasons I vote for removing that feature.

g-gemignani:
@mgrrx would it make sense now that https://github.com/ros/ros_comm/pull/1515 has been merged to switch to get_param_cached?

g-gemignani:
closing due to https://github.com/ros/diagnostics/commit/2e2e4ba3a5829ee59f97f6dbf753fed9def757c2

PR #87, opened 2018-09-27
Added wireless monitor diagnostic node

reinzor:
Using `sysstat` and `iwconfig`


trainman419:
We've been trying to keep this package generic so that it can be used on Linux and Mac. It's a core ROS package, so any breakage is pretty significant ( see https://github.com/ros/diagnostics/issues/36 for the last time this happened).

I suggest you submit this to https://github.com/ros-drivers/linux_peripheral_interfaces instead, since that package is explicitly for linux-specific system and hardware monitoring.

reinzor:
Thanks for the reply: clear. Are you ok with introducing a `linux_common_diagnostics` package in that repo?

reinzor:
Putting this into private repo

PR #88, opened 2018-10-02
/diagnostic_updater/DiagnosticTaskVector: added virtual destructor

cwecht:
This fixes a 'has virtual functions and accessible non-virtual destructor'-warning.

### DIFF
--- a/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.h
+++ b/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.h
@@ -240,7 +240,8 @@ namespace diagnostic_updater
         return tasks_;
       }
 
-    public:    
+    public:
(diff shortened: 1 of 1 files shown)


mikepurvis:
👍

trainman419:
As awesome as this is, it's also an ABI-breaking change. I'll probably need to hold it until the next major ROS release.

g-gemignani:
Thank you for the PR!

Unfortunately I don't think we should merge this one.

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160).
Every change comes with a risk of introducing regressions, and there isn't much time left to fix them.
I'm closing pull requests that add features so that the remaining time is allocated towards bug fixes and compatibility with newer Ubuntu distros.

PR #95, opened 2019-02-08
use operators instead of aliases

kejxu:
it is pretty straight-forward to just use the operators instead of alias

to access alias for logical `and`/`or`, extra setup is needed (https://docs.microsoft.com/en-us/cpp/cpp/logical-and-operator-amp-amp?view=vs-2017); otherwise, it would cause build break


trainman419:
Thanks!

PR #97, opened 2019-02-13
fix windows build issue

kejxu:
[`finite`](https://linux.die.net/man/3/finite) is Linux-only, it's [`_finite`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/finite-finitef?view=vs-2017) on Windows
to work around this, use [`isfinite`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/finite-finitef?view=vs-2017) on Windows, this also seems to exist on Linux (https://linux.die.net/man/3/isfinite)

`ERROR` is defined in windows.h, need to undefine the macro to fix build failure

### DIFF
--- a/diagnostic_updater/src/example.cpp
+++ b/diagnostic_updater/src/example.cpp
@@ -35,6 +35,8 @@
 #include <diagnostic_updater/diagnostic_updater.h>
 #include <std_msgs/Bool.h>
 #include <diagnostic_updater/publisher.h>
+
+// ERROR defined in windows.h causes name collision, undefine the macro to fix the issue
 #ifdef ERROR
 #undef ERROR
 #endif
--- a/diagnostic_updater/src/example.cpp
+++ b/diagnostic_updater/src/example.cpp
@@ -35,6 +35,9 @@
 #include <diagnostic_updater/diagnostic_updater.h>
 #include <std_msgs/Bool.h>
 #include <diagnostic_updater/publisher.h>
+#ifdef ERROR
+#undef ERROR
+#endif
 
 double time_to_launch;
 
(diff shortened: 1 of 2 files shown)


diagnostic_updater/src/example.cpp
```diff
@@ -36,6 +36,11 @@
 #include <std_msgs/Bool.h>
 #include <diagnostic_updater/publisher.h>
 
+// ERROR defined in windows.h causes name collision, undefine the macro to fix the issue
+#ifdef ERROR
+#undef ERROR
+#endif
```
trainman419, line 42: This may work here, but it indicates a problem with the message definition, and every other user of these messages will have the same issue. Is there some other way to prevent this, in a way that would apply to all of the users of this library?

kejxu: another method that might sound better and more universal (but actually bad) is to move similar macro guard into where `.msg` is generated into `.h`. however, there are 2 problems with this:
- the (centralized) generator would have no idea of the names that each `.msg` would use, it can never prevent all possible name collisions
- if all similar macros are un-defined, it might be doing extra unnecessary work, and actually break code that uses such macros

based on these, the very likely best (at this point) way is still just to handle these special cases right at where it happens =)

kejxu: ROS2 is (at this point) recommending similar fix to this problem, https://github.com/ros2/common_interfaces/issues/44#issuecomment-397516076

(thanks to @seanyen-msft for finding this)

kejxu:
@trainman419, just a friendly ping =) any update on this change?

diagnostic_updater/src/example.cpp
trainman419: Ok.

PR #101, opened 2019-03-06
ament ignores in all packages

Karsten1987:
Base PR for disabling the compilation of every package within the diagnostics ros2 branch.

This allows follow up PRs to be stacked upon each other and re-enabling a single package ready to be migrated and thus tested.


trainman419:
Awesome!

PR #102, opened 2019-03-06
Ros2 migrate diagnostic updater

Karsten1987:
Connects to ros/diagnostics#101

PR migrates `diagnostic_updater`. 

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6321/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5202/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6171/)

The failing tests are due to the ros2 buildfarm not being provisioned with `httplib2`. Tests passing locally though.


diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
```diff
@@ -305,36 +324,41 @@ def setHardwareID(self, hwid):
...
-        now = rospy.Time.now()
-        if  now >= self.last_time_period_checked + rospy.Duration(0.25):
+        clock = Clock(clock_type=ClockType.STEADY_TIME)
+        now = clock.now()
+        if now >= self.last_time_period_checked:
             try:
-                self.period = rospy.get_param("~diagnostic_period", 1)
                 self.last_time_period_checked = now
-            except (httplib.CannotSendRequest, httplib.ResponseNotReady):
+            except (httplib2.CannotSendRequest, httplib2.ResponseNotReady):
```
trainman419: Since the param call here has been removed, it looks like the try-except can be removed here too, and the dependency on httplib2 can be removed completely. (This should also fix the CI failure).

diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
```diff
@@ -206,90 +217,98 @@ def removeByName(self, name):
...
 
-    def __init__(self):
-        """Constructs an updater class."""
+    def __init__(self, node):
+        """Construct an updater class."""
         DiagnosticTaskVector.__init__(self)
-        self.publisher = rospy.Publisher("/diagnostics", DiagnosticArray, queue_size=10)
+        self.node = node
+        self.publisher = self.node.create_publisher(DiagnosticArray, '/diagnostics')
+        clock = Clock(clock_type=ClockType.STEADY_TIME)
```
trainman419: Instead of constructing a `Clock` every time one is needed in this class, does it make more sense to create it in the constructor and store it in a member variable?

trainman419: Instead of constructing a `Clock` every time one is needed in this class, does it make more sense to create it once in the constructor and store it in a member variable?

diagnostic_updater/include/diagnostic_updater/DiagnosticStatusWrapper.hpp
```diff
@@ -164,10 +166,11 @@ class DiagnosticStatusWrapper : public diagnostic_msgs::DiagnosticStatus
   void summaryf(unsigned char lvl, const char * format, ...)
   {
     va_list va;
-    char buff[1000];     // @todo This could be done more elegantly.
+    char buff[1000];  // @todo This could be done more elegantly.
     va_start(va, format);
-    if (vsnprintf(buff, 1000, format, va) >= 1000) {
-      ROS_DEBUG("Really long string in DiagnosticStatusWrapper::addf, it was truncated.");
+    if (vsnprintf(buff, sizeof(buff), format, va) >= 1000) {
```
trainman419: If this is going to change, it can be ` if (vsnprintf(buff, sizeof(buff), format, va) >= sizeof(buff) {`

diagnostic_updater/include/diagnostic_updater/DiagnosticStatusWrapper.hpp
```diff
@@ -257,19 +258,21 @@ inline void DiagnosticStatusWrapper::add<bool>(const std::string & key, const bo
...
+  const std::string & key, const char * format,
+  ...)                              // In practice format will always be a char *
 {
   va_list va;
-  char buff[1000];   // @todo This could be done more elegantly.
+  char buff[1000];  // @todo This could be done more elegantly.
   va_start(va, format);
-  if (vsnprintf(buff, 1000, format, va) >= 1000) {
-    ROS_DEBUG("Really long string in DiagnosticStatusWrapper::addf, it was truncated.");
+  if (vsnprintf(buff, sizeof(buff), format, va) >= 1000) {
```
trainman419: If this is going to change, it can be `if (vsnprintf(buff, sizeof(buff), format, va) >= sizeof(buff) {`

diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp
```diff
@@ -375,10 +372,12 @@ class Updater : public DiagnosticTaskVector
    * parameter.
    */
   Updater(
-    ros::NodeHandle h = ros::NodeHandle(), ros::NodeHandle ph = ros::NodeHandle("~"),
-    std::string node_name = ros::this_node::getName())
+    rclcpp::Node::SharedPtr h = rclcpp::Node::make_shared("diagnostics_updater"),
+    rclcpp::Node::SharedPtr ph = rclcpp::Node::make_shared("diagnostics_updater_private"),
```
trainman419: Does it really make sense to create two new ROS nodes here? I'm not too familiar with ROS2, but I think this actually creates a new node.

diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp
```diff
@@ -504,19 +509,17 @@ class Updater : public DiagnosticTaskVector
   void setHardwareIDf(const char * format, ...)
   {
     va_list va;
-    char buff[1000];     // @todo This could be done more elegantly.
+    char buff[1000];  // @todo This could be done more elegantly.
     va_start(va, format);
-    if (vsnprintf(buff, 1000, format, va) >= 1000) {
-      ROS_DEBUG("Really long string in diagnostic_updater::setHardwareIDf.");
+    if (vsnprintf(buff, sizeof(buff), format, va) >= 1000) {
```
trainman419: `if (vsnprintf(buff, sizeof(buff), format, va) >= sizeof(buff))`

trainman419:
According to https://index.ros.org/doc/ros2/Tutorials/Logging-and-logger-configuration/ , there are logging macros in rclcpp (like `RCLCPP_INFO`). Please use those instead of removing the log statements.

diagnostic_updater/diagnostic_updater/example.py
```diff
@@ -223,26 +225,31 @@ def check_upper_bound(stat):
...
+    while rclpy.ok():
         msg = std_msgs.msg.Bool()
-        rospy.sleep(0.1)
+        sleep(0.1)
 
         # Calls to pub1 have to be accompanied by calls to pub1_freq to keep
         # the statistics up to date.
         msg.data = False
         pub1.publish(msg)
+        pub2_temp.publish(msg)
```
Karsten1987: do you know what's the deal with this `pub2_temp` publisher instance?

diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
Karsten1987: I just added the parameters back. However, I am not familiar with the ROS1 python API well enough to judge whether the `httplib2` is the right type of exception here. Do you think we can replace that with some other type of exception to get rid of the `httplib2` dependency?

diagnostic_updater/diagnostic_updater/example.py
```diff
@@ -188,11 +190,10 @@ def check_upper_bound(stat):
 
     # You can broadcast a message in all the DiagnosticStatus if your node
     # is in a special state.
-    updater.broadcast(0, "Doing important initialization stuff.")
+    updater.broadcast(b'0', 'Doing important initialization stuff.')
 
-    pub1 = rospy.Publisher("topic1", std_msgs.msg.Bool, queue_size=10)
-    pub2_temp = rospy.Publisher("topic2", std_msgs.msg.Bool, queue_size=10)
```
Karsten1987, line 194: I don't know exactly what's the purpose of this `pub2_temp` publisher, but it seemed completely unused. Therefore, I went ahead and deleted it.
Same on the c++ API.

diagnostic_updater/include/diagnostic_updater/DiagnosticStatusWrapper.hpp
```diff
@@ -140,10 +145,13 @@ class DiagnosticStatusWrapper : public diagnostic_msgs::DiagnosticStatus
   void mergeSummaryf(unsigned char lvl, const char * format, ...)
   {
     va_list va;
-    char buff[1000];     // @todo This could be done more elegantly.
+    const int kBufferSize = 1000;
```
Karsten1987, line 148: that was the only way I've seen so far which seemed to make the ros2 linters happy. Hope that's okay for you as well.

diagnostic_updater/src/example.cpp
```diff
@@ -184,9 +186,8 @@ int main(int argc, char ** argv)
   // is in a special state.
   updater.broadcast(0, "Doing important initialization stuff.");
 
-  ros::Publisher pub1 = nh.advertise<std_msgs::Bool>("topic1", 1);
-  ros::Publisher pub2_temp = nh.advertise<std_msgs::Bool>("topic2", 1);
```
Karsten1987, line 188: as mentioned above, I also went ahead and deleted this unused variable.

diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
trainman419: Yeah; I think this was intended to catch errors if the parameter retrieval failed. I'm familiar with ROS2, but it seems like this should be replaced by a more appropriate `rclpy` exception (or completely removed, if param retrieval cannot fail in `rclpy`)

Karsten1987: So I removed the exception handling all together as I believe the `rclpy` implementation doesn't throw any errors:
https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/node.py#L231-L240

The only thing which could happen is that the parameter is not set. However, as we initialize the parameter ourselves in the constructor, this can't happen.

Karsten1987:
This is ready for re-review.  Hopefully all comments were addressed and the buildfarm seems happy with it as well.

I rebased the commits to have a minimum of semantically separated sets to review.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6432/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2810/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5285/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6259/)

diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
trainman419: Great!

PR #96, opened 2019-02-13
update CMakeLists.txt to search for local gtest first

kejxu:
gtest `1.7.0` on Windows causes the following error:
```
C:\ros\catkin_ws\upstream\ros-desktop\src\diagnostics\diagnostic_aggregator\gtest-1.7.0\gtest-all.cc(1374): error C2011: 'testing::internal::AutoHandle': 'class' type redefinition
C:\opt\rosdeps\x64\include\gtest/internal/gtest-port.h(1511): note: see declaration of 'testing::internal::AutoHandle'
C:\ros\catkin_ws\upstream\ros-desktop\src\diagnostics\diagnostic_aggregator\gtest-1.7.0\gtest-all.cc(3381): error C2440: 'delete': cannot convert from 'const testing::internal::scoped_ptr<testing::internal::GTestFlagSaver>' to 'void*'
C:\ros\catkin_ws\upstream\ros-desktop\src\diagnostics\diagnostic_aggregator\gtest-1.7.0\gtest-all.cc(3381): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
C:\ros\catkin_ws\upstream\ros-desktop\src\diagnostics\diagnostic_aggregator\gtest-1.7.0\gtest-all.cc(3667): error C2511: 'testing::TestInfo::TestInfo(const std::string &,const std::string &,const char *,const char *,testing::internal::TypeId,testing::internal::TestFactoryBase *)': overloaded member function not found in 'testing::TestInfo'
C:\opt\rosdeps\x64\include\gtest/gtest.h(644): note: see declaration of 'testing::TestInfo'
C:\ros\catkin_ws\upstream\ros-desktop\src\diagnostics\diagnostic_aggregator\gtest-1.7.0\gtest-all.cc(3676): error C2550: 'testing::TestInfo::{ctor}': constructor initializer lists are only allowed on constructor definitions
C:\ros\catkin_ws\upstream\ros-desktop\src\diagnostics\diagnostic_aggregator\gtest-1.7.0\gtest-all.cc(3710): error C2661: 'testing::TestInfo::TestInfo': no overloaded function takes 6 arguments
C:\ros\catkin_ws\upstream\ros-desktop\src\diagnostics\diagnostic_aggregator\gtest-1.7.0\gtest-all.cc(3709): error C2789: 'test_info': an object of const-qualified type must be initialized
[...]

PR #103, opened 2019-03-11
Reverse compatibility

twdragon:
Added reverse compatibility for newer Boost libraries and non-standard systems like SPARC (tested on SPARC Embedded). Compilation ability obtained for systems on which Google.Test framework installed with version that differs from default 1.7.0 bundled within the package.

### DIFF
--- a/diagnostic_aggregator/CMakeLists.txt
+++ b/diagnostic_aggregator/CMakeLists.txt
@@ -20,7 +20,7 @@ if( (${diagnostic_msgs_VERSION} VERSION_EQUAL "1.12.0") OR
 endif()
 
 find_package(Boost REQUIRED COMPONENTS system)
-include_directories(include ${catkin_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS} gtest-1.7.0/include)
+include_directories(include gtest-1.7.0/include ${catkin_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS})
 
 add_library(${PROJECT_NAME}
   src/status_item.cpp
@@ -29,7 +29,7 @@ add_library(${PROJECT_NAME}
   src/discard_analyzer.cpp
   src/ignore_analyzer.cpp
   src/aggregator.cpp)
-target_link_libraries(diagnostic_aggregator ${Boost_LIBRARIES}
+target_link_libraries(${PROJECT_NAME} ${Boost_LIBRARIES}
                                             ${catkin_LIBRARIES}
 )
 add_dependencies(${PROJECT_NAME} diagnostic_msgs_generate_messages_cpp)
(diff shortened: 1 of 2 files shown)


diagnostic_aggregator/CMakeLists.txt
```diff
@@ -38,12 +38,13 @@ add_dependencies(${PROJECT_NAME} diagnostic_msgs_generate_messages_cpp)
...
 target_link_libraries(aggregator_node ${catkin_LIBRARIES}
                                       ${PROJECT_NAME}
+                                      ${Boost_INCLUDE_DIRS}
 )
 
 # Analyzer loader allows other users to test that Analyzers load
 add_executable(analyzer_loader test/analyzer_loader.cpp
                                gtest-1.7.0/gtest-all.cc)
-target_link_libraries(analyzer_loader diagnostic_aggregator)
+target_link_libraries(analyzer_loader ${PROJECT_NAME} ${Boost_INCLUDE_DIRS})
```
trainman419, line 47: Why is this linking against `${Boost_INCLUDE_DIRS}` ?

trainman419:
Can you explain why the changes the to the CMakeLists.txt are necessary? To me it just seems like the path ordering has changed, and I didn't think that mattered in this situation. Is there something special about SPARC that's sensitive to include or linking order?

diagnostic_aggregator/CMakeLists.txt
twdragon: @trainman419 building the package on our non-standard SPARC embedded device we encountered an `undefined reference` error for Boost_regexp functions on `libdiagnostics_aggregator.so` without this directive. This is strange but adding the link solved the problem.

twdragon:
> Can you explain why the changes the to the CMakeLists.txt are necessary? To me it just seems like the path ordering has changed, and I didn't think that mattered in this situation. Is there something special about SPARC that's sensitive to include or linking order?

Yes it is sensitive to linking order, so if we have e.a. Google.Test framework installed onto the system, it interferes with the bundled one if linking order left unchanged.

diagnostic_aggregator/CMakeLists.txt
trainman419: If you have undefined references to boost, I suggest you try `${Boost_LIBRARIES}` here instead of `${Boost_INCLUDE_DIRS}`.

trainman419:
I suggest you try the changes from https://github.com/ros/diagnostics/pull/96 (just merged) and see if that resolves your issues with gtest.

diagnostic_aggregator/CMakeLists.txt
twdragon: Damn, it is my mistake here, not on the embedded machine! It was an uncorrect copypaste when I reproduced the changes on local repository by hand... Sorry!.. Of course, LIBRARIES should be here.

g-gemignani:
Closing due to no recent comments and fix already in https://github.com/ros/diagnostics/pull/96

PR #105, opened 2019-03-25
Fix httplib import for Python 3.

mikepurvis:
Per the official docs:

> The `httplib` module has been renamed to http.client in Python 3. The 2to3 tool will automatically adapt imports when converting your sources to Python 3.

https://docs.python.org/2/library/httplib.html#module-httplib


trainman419:
Thanks!

PR #108, opened 2019-05-14
Resolve div by zero in updater

### LINKED ISSUE #92: Division by zero if time does not change between updates
To determine the update frequency, the diagnostic_updater divides by the time since the last update here: https://github.com/ros/diagnostics/blob/indigo-devel/diagnostic_updater/include/diagnostic_updater/update_functions.h#L174

If the time has not changed since then (happend to me when playing a rosback with `use_sim_time`), we divide by zero.

wkhudgins92:
This resolves https://github.com/ros/diagnostics/issues/92. When the time window is 0 the program fails due to division by 0. With this change, division by 0 will not occur and instead the frequency will be omitted from the stats.


wkhudgins92:
Not sure why build is failing...doesn't seem to be related to my code

g-gemignani (approved):
Thank you for the improvement!

PR #113, opened 2019-06-22
Use std::isfinite due to deprecation on osx

sundermann:
On osx the c function finite is deprecated resulting in a build failure because clang fails on deprecated functions by default.

std::isfinite is in C++11 so the specific hadling for windows can be removed here as well.


g-gemignani (approved):
Thank you for the patch!

PR #114, opened 2019-06-24
Use timer

Karsten1987:
This is currently based on #112, but thinking about it, I believe it makes a lot of sense to use the inbuilt timer functionalities of ROS2 instead of manually comparing time stamps. 
This eases a lot of the logic and makes it easier to read as well.

However, this breaks API. But given that this package hasn't been released yet, I believe that's okay.


Karsten1987:
CI for this:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7425/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3575/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6092/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7247/)

Karsten1987:
new round of CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8014/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4026/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6525/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7865/)

Karsten1987:
@norro @ralph-lange Do you mind reviewing this?

Karsten1987:
windows warnings addressed:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8015/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4027/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6526/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7868/)

diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
```diff
@@ -229,40 +229,22 @@ class Updater(DiagnosticTaskVector):
...
         self.warn_nohwid_done = False
 
     def update(self):
         """Causes the diagnostics to update if the inter-update interval has been exceeded."""
-        self._check_diagnostic_period()
-        now = self.clock.now()
-        if now >= self.last_time:
-            self.force_update()
-
-    def force_update(self):
```
norro, line 257: Did the timer refactoring make it necessary to drop force_update (in python and c++)?

diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
```diff
@@ -229,40 +229,22 @@ class Updater(DiagnosticTaskVector):
     reason.
```
norro: If force_update is dropped, documentation block needs to be updated, as it still mentions this feature.

diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp
```diff
@@ -369,35 +361,44 @@ class Updater : public DiagnosticTaskVector
   /**
```
norro, line 357: Same as with python: if force_udpate is kicked out, documentation block needs to be updated.

norro (commented):
I like that you are now using the built-in timer. I stumbled upon kicking out `force_update`, but I guess, this is due to the timer change. Other than that, it looks good to me. 

I couldn't run the code, because I seem to need a newer rclcpp version than the one I have on my machine (for `create_timer`). But this is checked by CI anyway, I suppose.

diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
Karsten1987: the update function didn't really do anything if the period wasn't exceeded yet. So I believe it doesn't make sense to have a `force_update` function, which basically then just calls update. 
If you wanted to force an update, you could always just call `update` on your diagnostic updater instance

Karsten1987:
Thanks for the review. I've addressed your concerns about documentation and some additional style mismatches in  c7efe7b 
Please give that another review :)

As for your missing `rclcpp` functions. That got addressed in a [patch release](https://github.com/ros2/rclcpp/pull/757), so you might have to run a `apt-get upgrade` to get the latest binaries of rclcpp.

diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
norro: My understanding of the original implementation was, that `force_update()` will always update, but `update()` will only call `force_update()` when the period was exceeded. See https://github.com/ros/diagnostics/blob/ros2-devel/diagnostic_updater/diagnostic_updater/_diagnostic_updater.py#L254-L255 

This would mean that the two methods indeed have a different behavior. Calling `update()` with twice the configured rate will _still_ only send updates in the configrued rate, calling `force_update()` with twice the configured rate will _actually_ send updates with twice the configured rate.

Karsten1987: I think I see what you mean. In https://github.com/ros/diagnostics/pull/114/commits/18f5c771a1ba8cd1b9374583cc0e28f09f6c2c70 I have re-introduced the `force_update` method to make it more explicit. The result of this is that the `update` function becomes private to avoid any confusion.

norro (approved):

Karsten1987:
final round of CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8047/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4054/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6554/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7914/)

PR #121, opened 2019-08-25
ROS2: Parameter description hotfix

klintan:
Couldn't compile in dashing because declare_parameter needed 3 arguments, haven't seen that before, but it seems the parameter_interface does not have a default value for ParameterDescription.

edit: fails because it runs on crystal, some of these changes are dashing-specific.

### DIFF
--- a/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp
+++ b/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp
@@ -393,7 +393,7 @@ public:
     warn_nohwid_done_(false)
   {
     double period = parameters_interface->declare_parameter(
-      "diagnostic_updater.period", rclcpp::ParameterValue(1.0)).get<double>();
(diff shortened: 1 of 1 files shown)


klintan:
Running tests locally everything is fine except for the parameter: name = 'diagnostic_updater.period'
even though it's not a topic it complains it's not a valid topic name. Should we change the parameter name to diagnostics_updated_period ?

Karsten1987:
The default value got backported to dashing in here https://github.com/ros2/rclcpp/pull/794

Are you sure you are working with the latest dashing release? You might have to upgrade your system (`sudo apt-get upgrade`) or try working with the latest testing release repos: https://wiki.ros.org/TestingRepository

Looking at the PR introducing the parameter, the build worked successfully on all platforms with dashing: https://github.com/ros/diagnostics/pull/112

I am going to close this as I believe using the latest dashing debs might solve your problem. Feel free to re-open this ticket if your problems persists.

PR #124, opened 2019-09-18
Fix GCC warnings with -Wpedantic

jacobperron:
GCC version 7.4.0


g-gemignani (approved):
Thank you for the PR!

PR #135, opened 2020-03-02
Fix DiagnosedPublisher

DensoADAS:
The DiagnosedPublisher should be a wrapper around a publisher for any type of message, but it could only be used with publishers for DiagnosticArray messages. This was fixed.

### DIFF
--- a/diagnostic_updater/include/diagnostic_updater/publisher.hpp
+++ b/diagnostic_updater/include/diagnostic_updater/publisher.hpp
@@ -46,6 +46,26 @@
 #include "rclcpp/publisher.hpp"
 #include "rclcpp/subscription.hpp"
 
+namespace
+{
+/**
+ * \brief Helper struct to check message for header member.
+ */
+template<typename T, typename = void>
+struct has_header : public std::false_type
+{};
+
+/**
+ * \brief Helper struct to check message for header member.
+ */
+template<typename T>
+struct has_header<T,
+  typename std::enable_if<std::is_same<std_msgs::msg::Header,
+  decltype(std::declval<T>().header)>::value>::type>
+  : public std::true_type
+{};
+}  // namespace
+
 namespace diagnostic_updater
 {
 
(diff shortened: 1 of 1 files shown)


diagnostic_updater/include/diagnostic_updater/publisher.hpp
```diff
@@ -207,10 +209,10 @@ class DiagnosedPublisher : public TopicDiagnostic
    * The timestamp to be used by the TimeStampStatus class will be
    * extracted from message.header.stamp.
    */
-  virtual void publish(const std::shared_ptr<T> & message)
+  virtual void publish(typename PublisherT::MessageUniquePtr message)
   {
     tick(message->header.stamp);
```
Karsten1987, line 237: I believe this might lead to segfaults if compiling with message types not containing a header

Karsten1987 (commented):
Thanks for the fix. Definitely appreciated. I have one remark though which I believe can easily be addressed with a bit of template logic. Essentially, by calling `tick` on the message headers, we have requirement on the message type to contain a header field.

I would appreciate if you could check that `MessageT` contains a header field within a `static_assert` in the constructor. Essentially something along the following lines should do it:

```c++
#include <type_traits>

struct WithHeader
{
	std::string header;
};

struct WithoutHeader
{};

template<class T, typename = void>
struct has_header : public std::false_type
{};

template<class T>
struct has_header<
	T,
	typename std::enable_if<
		std::is_same<std::string, decltype(declval<T>().header)>::value
	>::type> : public std::true_type
{};

template<class T>
void do_stuff_with_header(T & instance)
{
	static_assert(has_header<T>::value, "message type has to have a header");
	instance.header = "Hello Header";
}

int main() {
	WithHeader w_header;
	WithoutHeader wo_header;
	do_stuff_with_header(w_header);
	do_stuff_with_header(wo_header);  // won't compile
	
	return 0;
}
```

diagnostic_updater/include/diagnostic_updater/publisher.hpp
kfabian: I think it just won't compile, but with the assert we get a better error message.

Karsten1987 (approved):
Thanks for iterating with me over this.

diagnostic_updater/include/diagnostic_updater/publisher.hpp
mikepurvis: You could make it work with header and non-header messages using the `HasHeader` trait, but I'm not sure it's worth the bother:

http://wiki.ros.org/roscpp/Overview/MessagesTraits#Optional_Traits

PR #107, opened 2019-04-03
[Windows][indigo] Use <chrono> APIs to replace usleep for better portability

seanyen:
This problem manifests when do `run_tests` on Windows build. Some test cases are using `usleep` which doesn't exist on some system (e.g., Windows).

This change is to propose use `<chrono>` APIs to achieve the same functionality.

PR #138, opened 2020-04-01
[Windows][noetic] Use <chrono> APIs to replace usleep for better portability

seanyen:
This is the same pull request #107 but target `noetic-devel`.


seanyen:
cc @Karsten1987 @ahcorde

Karsten1987 (approved):

PR #139, opened 2020-05-15
Allow zero analyzers without error

lucasw:
I'm using the aggregator to see the diagnostics without analysis, and don't want an error message when it starts up.


g-gemignani:
I honestly do not see this as an issue and I would avoid changing the behavior

lucasw:
It looks like the solution is to put in a dummy analyzer then.

PR #140, opened 2020-05-21
Fix python positional argument

Tobias-Fischer:
```
  File "lib/python3.6/site-packages/diagnostic_common_diagnostics/sensors_monitor.py", line 220
    print(file=sys.stderr, 'Unable to initialize node. Master may not be running')
                          ^
SyntaxError: positional argument follows keyword argument
```


g-gemignani:
Already fixed in https://github.com/ros/diagnostics/pull/141/files Thank you for submitting the patch though!

PR #142, opened 2020-06-11
Update topic_monitor

seiya0412:
- Change updaters_ to std::unordered_map from std::vector in order to avoid generating redundant updater.
- Use round-robin style update()


seiya0412:
This was accidentally opened. Ignore it, please.

Issue #45, opened 2016-02-22
Topics published on a frequency lower than 1.0/diagnostic_period are not supported


peci1:
Looking at `TimeStampStatus` and `FrequencyStatus` code (and testing with our hardware), if you use the `TopicDiagnostic` to diagnose a slow topic (0.3 Hz in our case), then the diagnostic fails.

At the time windows a message has arrived, it correctly reports on the frequency and all of this, but at the other time windows when no message has appeared, the diagnostic reports ERROR with `No events recorded. No data since last update.`. 

This is obviously true, but it is not an error state.

peci1:
Here's my workaround (tested to work correctly).

https://gist.github.com/peci1/ede392a7006d5a9af6c8

In the workaround, you can also see the basic problem of the provided `TimeStampStatus` and `FrequencyStatus` classes - their members are defined `private` instead of `protected`, effectively disabling people to extend this class and make full use of it.

trainman419:
This looks useful, but I don't have a lot of time to integrate it. It will go faster if you submit it as a pull request.

g-gemignani:
Closing since the feature has been merged and will be added to both melodic and noetic with the next versions

peci1:
I was looking for a good place to improve documentation, but I haven't found any that would specifically cope with FrequencyStatus and TimeStampStatus. Is there such a place that I overlooked?

g-gemignani:
Hi @peci1 and thank you for your message. If I am not wrong, in the past the diagnostics maintainers have added documentation to the wiki: http://wiki.ros.org/diagnostics I believe that would be the best place where to add info about your change

peci1:
I finally found time to put this into the docs! http://wiki.ros.org/action/info/diagnostic_updater?action=diff&rev2=26&rev1=25

g-gemignani:
Thank you for your time!

PR #144, opened 2020-06-26
Better support for slow topics

peci1:
This PR provides a solution for #45.

It adds a test for FrequencyStatus for the case when a topic is published on a slower rate than 1/diagnostic_period.

For TimeStampStatus, it adds a parameter `no_data_is_problem_` which allows to specify whether an "empty run()" should be treated as error (for slow topics, it is okay when there are updates without any data).

The last commit is an ABI break (API remains compatible). I needed to add the field to remember whether to treat no data as error or not. There are ways to get around the ABI break, but none of them is good (either a global static map<TimeStampStatus, bool> with a lock, or "misusing" some higher bits in one of the private parameters of TimeStampStatus).

### DIFF
--- a/diagnostic_updater/include/diagnostic_updater/update_functions.h
+++ b/diagnostic_updater/include/diagnostic_updater/update_functions.h
@@ -227,7 +227,11 @@ namespace diagnostic_updater
      */
 
     TimeStampStatusParam(const double min_acceptable = -1, const double max_acceptable = 5) :
-      max_acceptable_(max_acceptable), min_acceptable_(min_acceptable)
+      max_acceptable_(max_acceptable), min_acceptable_(min_acceptable), no_data_is_problem_(true)
+    {}
+
+    TimeStampStatusParam(const double min_acceptable, const double max_acceptable, const bool no_data_is_problem) :
+      max_acceptable_(max_acceptable), min_acceptable_(min_acceptable), no_data_is_problem_(no_data_is_problem)
     {}
 
     /**
@@ -242,6 +246,10 @@ namespace diagnostic_updater
 
     double min_acceptable_;
 
+    /**
+     * \brief Whether to report a warning when no data arrived since last update.
+     */
+    bool no_data_is_problem_;
   };
 
   /**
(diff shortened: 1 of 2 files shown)


peci1:
I also took the opportunity to change the DiagnosticUpdater tests to use simulated time, which should remove any possible flakyness from them.

peci1:
Travis CI is broken (unrelated to this PR):

```
$ sudo apt-get install python-catkin-pkg python-rosdep ros-indigo-catkin
...
Selecting previously unselected package cmake-data.

Preparing to unpack .../cmake-data_2.8.12.2-0ubuntu3_all.deb ...

ERROR: cmake-data is broken - called emacs-package-install as a new-style add-on, but has no compat file.
```

peci1:
On build.ros.org, there's one test failure in `__main__.TestDiscardStale.test_discard_stale`, but that one should not be related at all...

peci1:
`test_discard_stale` was flaky and that was fixed in #145.

peci1:
I finally found a non-hacky ABI-compatible way to add this functionality. I created a "tagging" subclass of `TimeStampStatus` which is then checked via dynamic_cast in the `run()` method. It'd be nicer to do the same with `TimeStampStatusParam`, but the struct is not polymorphic, so dynamic_cast wouldn't work on it.

The dynamic_cast wouldn't be needed, as I could easily just re-implement the virtual `run()` method in the derived class, but I'm not a friend of copy-paste code. And any attempt to share the code between the base and derived class I can think of would break ABI.

When porting this PR to newer distros which can break ABI, I'd strongly suggest to not cherry-pick the ABI-compatibility-making commit 8d17. It makes much more sense to choose the behavior for slow topics in a parameter than by subclassing the diagnostics class.

g-gemignani (approved):
Thank you for your contribution, looks good to me! :)

peci1:
Do you intend to backport this to indigo-devel? Or should I prepare the PR?

g-gemignani:
I can take care of it. Thank you :)

peci1:
Great, thanks!

PR #145, opened 2020-06-28
diagnostic_aggregator: Fixed flaky test.

peci1:
discard_stale_not_published_test.py had an initial wait for /diagnostics topic and a subsequent wait for stale agg messages. If, however, the initial wait for /diagnostics took longer (e.g. high CPU utilization), it decreased the amount of time left to wait for stale messages. The fix makes these wait times independent.

Example build when the test failed: http://build.ros.org/job/Npr__diagnostics__ubuntu_focal_amd64/10


g-gemignani:
Thank you for your contribution!

PR #148, opened 2020-07-01
restore alphabetical order

Karsten1987:
The issue listed in https://github.com/ros2/ros2/issues/927 got resolved, which allows an alphabetical include order


norro (approved):
Looks good, works for me.

PR #152, opened 2020-07-13
feat(cpu_monitor): Window averaging

reinzor:
Added `~window` parameter in order to average multiple cpu percentage readings.

### DIFF
--- a/diagnostic_common_diagnostics/src/diagnostic_common_diagnostics/cpu_monitor.py
+++ b/diagnostic_common_diagnostics/src/diagnostic_common_diagnostics/cpu_monitor.py
@@ -34,35 +34,45 @@
 
 # \author Rein Appeldoorn
 
-import rospy
-from diagnostic_updater import DiagnosticTask, Updater
-from diagnostic_msgs.msg import DiagnosticStatus
+import collections
+import socket
 
 import psutil
-import socket
+import rospy
+from diagnostic_msgs.msg import DiagnosticStatus
+from diagnostic_updater import DiagnosticTask, Updater
 
 
 class CpuTask(DiagnosticTask):
-    def __init__(self, warning_percentage):
+    def __init__(self, warning_percentage, window):
         DiagnosticTask.__init__(self, "CPU Information")
         self._warning_percentage = int(warning_percentage)
+        self._readings = collections.deque(maxlen=window)
+
+    def _get_average_reading(self):
+        def avg(lst):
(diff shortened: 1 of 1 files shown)


reinzor:
Ping @g-gemignani

diagnostic_common_diagnostics/src/diagnostic_common_diagnostics/cpu_monitor.py
```diff
@@ -34,35 +34,45 @@
...
 class CpuTask(DiagnosticTask):
-    def __init__(self, warning_percentage):
+    def __init__(self, warning_percentage, window):
         DiagnosticTask.__init__(self, "CPU Information")
         self._warning_percentage = int(warning_percentage)
+        self._readings = collections.deque(maxlen=window)
+
+    def _get_average_reading(self):
+        def avg(lst):
+            return float(sum(lst)) / len(lst)
```
g-gemignani: Could you please also handle the len(lst) = 0 case just to be on the safe side?

g-gemignani (changes requested):
Would adding a test be too much to ask?

g-gemignani:
On second thought, @reinzor do you need this improvement also in indigo? I would otherwise just add this new feature into noetic

diagnostic_common_diagnostics/src/diagnostic_common_diagnostics/cpu_monitor.py
reinzor: Done

reinzor:
> On second thought, @reinzor do you need this improvement also in indigo? I would otherwise just add this new feature into noetic

It would be nice to have this in `Melodic` as well (it is backwards compatible and should not affect any running configuration). Since `Kinetic`/`Melodic` are using the `indigo-devel` branch that would be nice :).

reinzor:
Thanks btw for continuing to remain this repo!

reinzor:
> Would adding a test be too much to ask?

Regarding the test: the CPU monitor is already being tested here: https://github.com/ros/diagnostics/tree/indigo-devel/diagnostic_common_diagnostics/test/launch

g-gemignani:
Correct, but since you are adding a feature I believe we should extend the test :)

g-gemignani:
Yep, it is true that the analyzer is tested there but I was thinking about creating another one where we set the window parameter to something different than 1, but I guess it is not needed.

g-gemignani (approved):

PR #153, opened 2020-07-13
feat(cpu_monitor): Window averaging

reinzor:
Same as #152 but then for `Noetic`


reinzor:
Ping @g-gemignani

g-gemignani:
Please see comments in https://github.com/ros/diagnostics/pull/152

PR #157, opened 2020-08-01
Instal example publisher for diagnostic_aggregator example as a program.

airballking:
Fixes https://github.com/ros/diagnostics/issues/156

PR #158, opened 2020-08-04
use ubuntu 20.04 for GH actions

Karsten1987:
This should hopefully address the concerns raised in [here](https://github.com/ros/diagnostics/pull/157#issuecomment-668646487)


Karsten1987:
checks are passing now. I'll go ahead and merge to unblock further incoming PRs.

Issue #156, opened 2020-08-01
[ROS2 | Foxy] diagnostic_aggregator example crashes because example_pub.py is not installed as a program


airballking:
It seems the example publisher of the diagnostic_aggregator publisher is not installed as a program. Hence, the overall example fails.

I ran this on ROS2 Foxy installed from Ubuntu version: 20.04 deb files. Last update was August 1st, 2020.

Here is my error description:

```shell
$ ros2 launch diagnostic_aggregator example.launch.py
[INFO] [launch]: All log files can be found below /home/myuser/.ros/log/2020-08-01-13-22-57-725483-mypc-30482
[INFO] [launch]: Default logging verbosity is set to INFO
/opt/ros/foxy/share/diagnostic_aggregator/example.launch.py:9: UserWarning: The parameter 'node_executable' is deprecated, use 'executable' instead
  aggregator = launch_ros.actions.Node(
/opt/ros/foxy/share/diagnostic_aggregator/example.launch.py:14: UserWarning: The parameter 'node_executable' is deprecated, use 'executable' instead
  diag_publisher = launch_ros.actions.Node(
[ERROR] [launch]: Caught exception in launch (see debug for traceback): executable 'example_pub.py' not found on the libexec directory '/opt/ros/foxy/lib/diagnostic_aggregator' 
Task exception was never retrieved
future: <Task finished name='Task-8' coro=<LaunchService._process_one_event() done, defined at /opt/ros/foxy/lib/python3.8/site-packages/launch/launch_service.py:271> exception=RuntimeError('Signal event received before subprocess transport available.')>
Traceback (most recent call last):
  File "/opt/ros/foxy/lib/python3.8/site-packages/launch/launch_service.py", line 273, in _process_one_event
    await self.__process_event(next_event)
  File "/opt/ros/foxy/lib/python3.8/site-packages/launch/launch_service.py", line 293, in __process_event
    visit_all_entities_and_collect_futures(entity, self.__context))
  File "/opt/ros/foxy/lib/python3.8/site-packages/launch/utilities/visit_all_entities_and_collect_futures_impl.py", line 38, in visit_all_entities_and_collect_futures
    sub_entities = entity.visit(context)
  File "/opt/ros/foxy/lib/python3.8/site-packages/launch/action.py", line 108, in visit
    return self.execute(context)
  File "/opt/ros/foxy/lib/python3.8/site-packages/launch/actions/opaque_function.py", line 75, in execute
[...]

PR #159, opened 2020-08-05
Fix example pub install ros2 devel

airballking:
Replaces https://github.com/ros/diagnostics/pull/157

Fixes https://github.com/ros/diagnostics/issues/156

@Karsten1987 Corrections as discussed in the original pull request.


Karsten1987:
thanks for the patch.

airballking:
You're welcome.

PR #173, opened 2021-01-05
Fix some doc typos and add industrial_ci

mikaelarguedas:
The first 2 commits fix some trivial documentation typos.
https://github.com/ros/diagnostics/commit/ddcaf12fbb854415ed2de1d48f1bad3da2fa6d43 configures a github actions workflow to test this branch against noetic using the industrial_ci utility.
https://github.com/ros/diagnostics/commit/4fc494aa95022646e425dfe7c4b5ef8f7074f7e3 removes the travis config in favor of the github actions one.


g-gemignani:
Hi @mikaelarguedas and thank you for your PR. As I am not familiar with industrial_ci, could you give me a bit of context on why you are proposing this switch? I know that this ci solution is also being discussed in the moveit repo (https://github.com/ros-planning/moveit/issues/2462) but afaik there isn't yet an agreement on its adoption or not. Thank you for the clarification! :)

mikaelarguedas:
>  could you give me a bit of context on why you are proposing this switch?

The main driver for me to suggest a change is that travis was failing on this PR (and apparently on master for a while) and it seemed to be more effort to fix the travis config as it looks to be a custom script. `industrial_ci` allows a generic way to test any repository against various ROS distributions. So it was the path of least resistance to get CI passing on this PR.

I'm not very familiar with the moveit(2) or navigation2 CI solution. My understanding is that these are projects that have a significant amount of code so installing dependencies and building is very time consuming. In addition they (some?) attempt to test against the latest ROS 2 which is something more challanging than testing against a released ROS distribution.

---


Arguably as there is already the `ros_buildfarm` PR jobs enabled on this repo there is maybe no need for a second CI infrastructure testing the same code ?

Karsten1987:
Just fyi, I've just recently updated the GH actions for the ROS2 Foxy branch to the ros-tooling action infrastructure:
https://github.com/ros/diagnostics/blob/foxy/.github/workflows/test.yaml

might be worth to keep it consistent across the repo. But there's nothing wrong with the industrial ci.

g-gemignani:
@mikaelarguedas I would go with Karsten's proposal. Would it be too much to ask to please modify the PR accordingly?

mikaelarguedas:
> I would go with Karsten's proposal. Would it be too much to ask to please modify the PR accordingly?

Last time I tried to use that for ROS 1 (~1year ago so things may have changed a lot) it proved harder to configure/maintain, less performant, (and a lot less battle-ready) than `industrial_ci`. It seems to me that that solution is strongly tailored to ROS 2 and offers limited usability for ROS 1 repositories.
For example I'm not sure how to configure jobs to run on all noetic platforms (Ubuntu Focal and Debian Buster) with that tool. Maybe someone more familiar with the tool (@Karsten1987 ?) can provide an example of such configuration?

In general I would say that `industrial_ci` is the goto for ROS 1 repositories as it has been designed for that purpose and is being used on thousands of repositories. And that the `action-ros-ci` is more useful for ROS 2 repositories that have specific needs (build all ROS 2 from source, test on MacOS/Windows, use specific colcon mixins etc) at the cost of verbosity and maintainability. 

This also brings me back to the previous question:
> Arguably as there is already the ros_buildfarm PR jobs enabled on this repo there is maybe no need for a second CI infrastructure testing the same code ?

Do you guys think having CI on this repo in addition to the PR jobs provided by ROS Buildfarm provides any additional value ?

g-gemignani:
@mikaelarguedas you have a point in questioning the need for a second CI infrastructure. Since I do not believe there will be much more development in this repository, I would then cut the discussion short and just stick with one CI infrastructure. Can you, therefore, remove the industrial_ci config? Thank you

mikaelarguedas:
> Can you, therefore, remove the industrial_ci config?

Sounds good, reverted in https://github.com/ros/diagnostics/pull/173/commits/c1751833073d7a994cb662d6e4c5355905c6f628

g-gemignani:
thank you!

PR #176, opened 2021-01-07
Update gh actions

Karsten1987:
as per title

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,7 +21,7 @@ jobs:
           required-ros-distributions: foxy
       - uses: ros-tooling/action-ros-ci@v0.1
         with:
-          source-ros-binary-installation: foxy
+          target-ros2-distro: foxy
           package-name: diagnostic_aggregator diagnostic_updater self_test
           colcon-mixin-name: coverage-gcc
           colcon-mixin-repository: https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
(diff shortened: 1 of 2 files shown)


Karsten1987:
@emersonknapp sorry for pinging you here with no context, but I am still struggling with creating an overlay workflow.
While I see with the actions designed that ros-foxy-desktop-full is getting installed, but then however in the ros-action-ci it's still fetching the master repos, effectively neglecting the binary installation:
```
Invoking "bash -c 'curl --fail --silent --show-error --location 'https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos' | vcs import --force --recursive src/'
```

What am I doing wrong in here? Thanks :)

.github/workflows/test.yaml
```diff
@@ -16,9 +16,12 @@ jobs:
...
         os: [ubuntu-20.04]
     steps:
-      - uses: ros-tooling/setup-ros@0.0.23
-      - uses: ros-tooling/action-ros-ci@0.0.17
+      - uses: ros-tooling/setup-ros@v0.1
         with:
+          required-ros-distributions: foxy
+      - uses: ros-tooling/action-ros-ci@v0.1
+        with:
+          target-ros2-distro: foxy
```
emersonknapp, line 24: This definitely isn't clear enough in the docs, and the default setting right now is probably the wrong experience.
If you want to build only against installed binary dependencies instead of checking out `ros2.repos`, you can specify an empty `vcs-repo-file-url` input parameter, which will override the default value of `https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos`

```suggestion
        with:
          target-ros2-distro: foxy
          vcs-repo-file-url: ""
```

Karsten1987: perfect! That seems to work. Thanks for the help.
And yes, that would be great to document, because I feel that's a valid use case for this GH action.

PR #179, opened 2021-01-19
[Foxy] Move Aggregator publishing to timer to allow subscription callback more processing time

cdbierl:
Moving the aggregator's publishing to a wall timer, and letting the node spin continuously, permits the subscription callback to operate at the necessary rate of incoming diagnostic entry messages.


Karsten1987:
This generally looks good to me. However, please be advised that this branch has to target `ros2-devel` and not `foxy`. We can't accept this patch in `foxy` as it's breaking API.

Also @norro you might want to weigh in here as the author of the aggregators ;-)

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -101,6 +102,10 @@ Aggregator::Aggregator()
   agg_pub_ = n_->create_publisher<diagnostic_msgs::msg::DiagnosticArray>("/diagnostics_agg", 1);
   toplevel_state_pub_ =
     n_->create_publisher<DiagnosticStatus>("/diagnostics_toplevel_state", 1);
+
+  int publish_rate_ms = 1000 / pub_rate_;
+  publish_timer_ = n_->create_wall_timer(std::chrono::milliseconds(publish_rate_ms),
```
norro: please uncrustify:
```
-  publish_timer_ = n_->create_wall_timer(std::chrono::milliseconds(publish_rate_ms),
+  publish_timer_ = n_->create_wall_timer(
+    std::chrono::milliseconds(publish_rate_ms),

File: diagnostic_aggregator/src/aggregator_node.cpp [hunk 2]
```diff
@@ -34,34 +34,20 @@
 
 /**< \author Kevin Watts */
 
-#include <exception>
+#include <rclcpp/rclcpp.hpp>
```
norro: Please include <memory> for using `make_shared`

norro (changes requested):
Thank you very much for this PR, it looks good to me as well!
Can you please address the two minor issues detected by cpplint resp. uncrustify and redirect this PR to ros2-devel?

diagnostic_aggregator/src/aggregator_node.cpp
cdbierl: Are you referring to the original try-catch?

norro: Sorry, the github syntax swallowed parts of the message.
I was asking to `#include <memory>` because you introduce `make_shared`.

cdbierl:
Any objections to me also bringing in the subscription history depth change ```.keep_last(1000)``` in the ROS2 PR ?

cdbierl:
> This generally looks good to me. However, please be advised that this branch has to target `ros2-devel` and not `foxy`. We can't accept this patch in `foxy` as it's breaking API.
> 
> Also @norro you might want to weigh in here as the author of the aggregators ;-)

What is the breaking API change you mention for Foxy?

norro:
@cdbierl `publishData()` seems to be private now after your PR.

cdbierl:
@norro That is the breaking API change? I also removed some other public functions no longer used in this package. Is the point that other people using the aggregator will find the breaking API change(s)? If you mean I have introduced local breaking API changes, then I understand. I can revise. I am not sure though if this is what @Karsten1987 meant.

Maybe I am coming around to the idea now. In Foxy, you aren't allowed to introduce breaking API changes, like I did by changing access level or removing API functions of the Aggregator class (annotated by the macros)? I suppose this is also the case for my ros2-devel PR? Let me know if I am on the right page now.

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -42,6 +42,7 @@
 #include <mutex>
 #include <string>
 #include <vector>
+#include <chrono>
```
Karsten1987: please sort the includes alphabetically.

diagnostic_aggregator/src/aggregator_node.cpp
```diff
@@ -34,34 +34,21 @@
...
 
-#include <exception>
-
 #include "diagnostic_aggregator/aggregator.hpp"
 
-#include "rclcpp/executors.hpp"
-
-using std::exception;
+#include <memory>
+#include <rclcpp/rclcpp.hpp>
```
Karsten1987:
```suggestion

#include "rclcpp/rclcpp.hpp"
```

Karsten1987 (approved):
Looks good to me with two nitpicks addressed and green CI (uncrustify, cppcheck)

Karsten1987:
> What is the breaking API change you mention for Foxy?

You have removed functions and moved some as private. That introduces a breaking API change where external users can't rely on these functions any longer within the same ros-distro.
It actually goes one step further where we don't want to even have ABI breaks, so that external users don't have to recompile their libraries using diagnostics in Foxy.
That is to say, new changes should always land on ros2-devel so that we don't have to care much about API/ABI breakages. If you want to use binary packages (debians) for the latest state of ROS2, we recommend to rely on the `rolling` distro.

> Any objections to me also bringing in the subscription history depth change .keep_last(1000) in the ROS2 PR ?

Let's keep the changes semantically separated. There's already a PR open for this and I'd rather push forward on that one than further diluting this PR.

norro (approved):
LGTM

diagnostic_aggregator/src/aggregator_node.cpp
cdbierl: Would you happen to know if there is a .clang-format file that follows your pipeline checks?

norro: I think you find it in the `ament_clang_format` package resp. in your ROS 2 workspace in `src/ament/ament_lint/ament_clang_format/ament_clang_format/configuration/.clang-format`

cdbierl:
Is there anything I need to do from the Rpr__diagnostics__ubuntu_focal_amd64 test failure? Is that trying to build this for Rolling?

Karsten1987:
~so I've changed the target branch to `ros2-devel` as discussed above. Do you might taking a look at the conflicts? I am confident that a simple `git rebase` should do the trick.~

My bad: Just realized that this PR deliberately targets `Foxy`. Please be advised then that the current status can't be merged due to API/ABI changes.

Karsten1987 (changes requested):
can't be merged due to API/ABI changes.

cdbierl:
@Karsten1987 I'm not sure I understand what needs to be adjusted. The only thing I see failing is this not building for Rolling in CI, but that has nothing to do with this changeset. What particularly is not valid?

Karsten1987:
> You have removed functions and moved some as private. That introduces a breaking API change where external users can't rely on these functions any longer within the same ros-distro.
It actually goes one step further where we don't want to even have ABI breaks, so that external users don't have to recompile their libraries using diagnostics in Foxy.

So you can't remove the two functions (API break) and you can't really add the timer base shared pointer to the header (ABI break). 
The KDE community has a quite helpful [website](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B) about compatibility.

Karsten1987 (approved):
that looks better now. Thanks for iterating with me over this.

PR #180, opened 2021-01-20
[ROS2] Move Aggregator publishing to timer to allow subscription callback more processing time

cdbierl:
Moving the aggregator's publishing to a wall timer, and letting the node spin continuously, permits the subscription callback to operate at the necessary rate of incoming diagnostic entry messages.


norro (approved):
LGTM

PR #181, opened 2021-01-21
[Foxy] Enable multiple tasks publishing for diagnostic updater

BasVolkers:
Currently, the publish() function of the Updater() class in _diagnostic_updater.py seems bugged.

A DiagnosticStatus is only created for the last `stat` in `msg`, while it should be for every `stat in msg`.
This means that only a single DiagnosticStatus is published instead of all the DiagnosticStatus.

This Pull Request fixes this by creating a DiagnositcStatus in the for loop and appending it to the DiagnosticArray.

PR #182, opened 2021-01-22
[ROS2] Enable multiple tasks publishing for diagnostic updater 

BasVolkers:
Currently, the publish() function of the Updater() class in _diagnostic_updater.py seems bugged.

A DiagnosticStatus is only created for the last stat in msg, while it should be for every stat in msg.
This means that only a single DiagnosticStatus is published instead of all the DiagnosticStatus.

This Pull Request fixes this by creating a DiagnositcStatus in the for loop and appending it to the DiagnosticArray.

This is a copy of https://github.com/ros/diagnostics/pull/181, but with target ros2-devel.

### DIFF
--- a/diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
+++ b/diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
@@ -329,18 +329,19 @@ class Updater(DiagnosticTaskVector):
         if not type(msg) is list:
             msg = [msg]
 
+        now = self.clock.now()
+        da = DiagnosticArray()
+        da.header.stamp = now.to_msg()  # Add timestamp for ROS 0.10
         for stat in msg:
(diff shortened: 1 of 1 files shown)


BasVolkers:
@Karsten1987

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13767/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8633/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11476/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13841/)

Karsten1987 (approved):

Karsten1987:
@BasVolkers thanks for the patch and sorry for the long turnaround. This ticket flew under my radar.

PR #184, opened 2021-01-29
[ROS2] Fix setting of diagnostic status level by diagnostic updater 

### LINKED ISSUE #183: [ROS2] [Bug] DiagnosticStatus level is not correctly set
If I run a `HeaderlessTopicDiagnostic` with a `FrequencyStatusParam` I get the following console output:
```
[aggregator_node-2] [ERROR] [1611923352.150835653] [generic_analyzer_base]: Attempting to convert 48 into DiagnosticLevel.
[aggregator_node-2]     Values are: {0: OK, 1: Warning, 2: Error, 3: Stale}
```

I'm thinking the problem lies in this piece of code (line 117-125 of `_update_functions.py`):
``` python
            if events == 0:
                stat.summary(b'2', 'No events recorded.')
            elif freq < self.params.freq_bound['min'] * (1 - self.params.tolerance):
                stat.summary(b'1', 'Frequency too low.')
            elif 'max' in self.params.freq_bound and freq > self.params.freq_bound['max'] *\
                 (1 + self.params.tolerance):
                stat.summary(b'1', 'Frequency too high.')
            else:
                stat.summary(b'0', 'Desired frequency met')
```
I think that the bytes for the `DiagnosticStatus` message are not correctly created, since 48 is the ASCII value of '0'

By changing line 125 to the following I was able to fix this error: 
``` python
                stat.summary(b'\x00', 'Desired frequency met')
```

I'm assuming the same should be applied every time a status summary is created?

BasVolkers:
This PR resolves #183 

It replaces every byte creation b'0' by b'\x00', b'1' by b'\x01', and b'2' by b'\x02'.
With these changes the diagnostic_aggregator_node no longer produces an ERROR.

### DIFF
--- a/diagnostic_updater/diagnostic_updater/_diagnostic_status_wrapper.py
+++ b/diagnostic_updater/diagnostic_updater/_diagnostic_status_wrapper.py
@@ -88,7 +88,7 @@ class DiagnosticStatusWrapper(DiagnosticStatus):
 
     def clearSummary(self):
         """Clear the summary, setting the level to zero and the message to."""
-        self.summary(b'0', '')
+        self.summary(b'\x00', '')
 
     def mergeSummary(self, *args):
         """
@@ -115,7 +115,7 @@ class DiagnosticStatusWrapper(DiagnosticStatus):
             lvl = args[0]
             msg = args[1]
 
-        if (lvl > b'0') == (self.level > b'0'):
+        if (lvl > b'\x00') == (self.level > b'\x00'):
             if len(self.message) > 0:
                 self.message += '; '
             self.message += msg
--- a/diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
+++ b/diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
@@ -249,7 +249,7 @@ class Updater(DiagnosticTaskVector):
         with self.lock:  # Make sure no adds happen while we are processing here.
             for task in self.tasks:
                 status = DiagnosticStatusWrapper()
-                status.level = b'2'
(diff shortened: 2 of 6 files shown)


BasVolkers:
@Karsten1987

Karsten1987 (approved):

Karsten1987:
@BasVolkers this looks to me as if we most likely don't test this correctly. Do you have an idea if we could enhance the test suite with some regression test for this change?

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13771/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8637/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11480/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13849/)

BasVolkers:
> @BasVolkers this looks to me as if we most likely don't test this correctly. Do you have an idea if we could enhance the test suite with some regression test for this change?

The tests only check that a certain value is set for the level of a DiagnosticStatus. Since the wrong values were used in the `expected` part of the assertEquals the tests still passed.

The actual error happens when the DiagnosticStatus's level is converted in `status_item.hpp`. 
Since b'0' gives the ASCII value of '0' (48), the conversion fails. This fix make sure the level is in range [0-3].
A better fix may be using the DiagnosticStatus.msg levels (OK, WARN, ERROR), instead of creating the byte by hand.

I'm opening up MR https://github.com/ros/diagnostics/pull/193 for this.

PR #187, opened 2021-02-23
Improve messages of GenericAnalyzer when items are missing.

MichaelGrupp:


MichaelGrupp:
Side note: noetic-devel is not the default branch, do I also need to commit to indigo-devel? Or can you set noetic-devel to the default branch in the GitHub repo settings?

g-gemignani:
Thanks for the Pr @MichaelGrupp can we please replace missing with stale?

MichaelGrupp:
changed

g-gemignani (approved):

PR #190, opened 2021-03-02
new launch file syntax

### LINKED ISSUE #189: [ros2, rolling] aggregator has outdated launch file
As of [this PR](https://github.com/ros2/launch_ros/pull/190/files) all keywords prefixed with `node_` are no longer deprecated but actually removed. 

that is, `node_executable` is only `executable` here: https://github.com/ros/diagnostics/blob/ros2-devel/diagnostic_aggregator/example/example.launch.py.in#L11

Karsten1987:
fixes #189 

all keywords in a launch file prefixed with `node_` are no longer deprecated but removed in rolling/galactic.


norro (approved):
Checked the example, all tests successfull.

PR #191, opened 2021-03-03
[backport] Replace every byte creation instance (#184)

Karsten1987:
Foxy backport PR for  #184.
/cc @basvolkers


norro (approved):
LGTM, tests/examples still working as expected.

PR #194, opened 2021-03-20
[diagnotics_analysis] fix type comparison for python3

mikaelarguedas:
In python3, `self._stats[name]['fields'].keys()` returns a `dict_keys` object, so the comparison to a `list` is never `True`.

PS: the default branch of this depo is `indigo-devel` making it the default target for PRs, should it be noetic-devel instead ?

### DIFF
--- a/diagnostic_analysis/src/diagnostic_analysis/exporter.py
+++ b/diagnostic_analysis/src/diagnostic_analysis/exporter.py
@@ -112,7 +112,7 @@ class LogExporter:
                 
 
             # Check to see if fields have changed. Add new fields to map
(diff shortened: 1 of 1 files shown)


g-gemignani:
Thanks for the PR.
About the default branch, unfortunately I do not have the rights to do so. Maybe @trainman419 can do that?

trainman419:
Ok; I've updated the default branch to `noetic-devel`

mikaelarguedas:
Awesome, thank you both for the prompt reply 👍

g-gemignani:
Thank you @trainman419

Issue #48, opened 2016-05-20
Publish diagnostics_agg right away upon new warn/error diagnostics


mikepurvis:
The present behaviour is to only publish the aggregated diagnostics at the fixed rate (default 1Hz):

https://github.com/ros/diagnostics/blob/03e0db006175c3e6157bb50ab38021ebb4995c5b/diagnostic_aggregator/src/aggregator_node.cpp#L50-56

I believe that it would be better to trigger an immediate publish of the aggregated topic when a diagnostic transitions to WARN or ERROR. We have some error reporting mechanisms that currently have to subscribe to `/diagnostics`, and could instead subscribe to `/diagnostics_agg` if we knew that a) no error reports would be missed, and b) new error reports would be passed through immediately.

Thoughts?

heuristicus:
This sounds like a good idea, but is there any particular advantage to working with `/diagnostics_agg` rather than `/diagnostics`? I guess the main thing would be fewer callbacks, and the filtering that you define through the analyzers would be applied so you only see the diagnostics you're interested in.

At the same time, it seems to me that the main purpose of the `/diagnostics_agg` right now is to provide information to a GUI, which is why the update rate is slow.

mikepurvis:
The main consumer of `diagnostics_agg` is the GUI, but I think its _purpose_ is to supply a single diagnostic snapshot which doesn't require opening a socket to every node which produces diagnostics.

Even for the GUI case though, IMO a new ERROR case should be made visible as soon as possible, rather than waiting as much as a second for the next report to arrive.

heuristicus:
Definitely agree with making errors visible ASAP.

What about changing messages that are attached to the diagnostics? Does it matter that the cause of the error is different, but the level is still the same?

For example:

```
level: 2
name: ''
message: something went wrong...
hardware_id: ''
values: 
  - 
    key: ''
    value: ''
---
level: 2
name: ''
message: something else went wrong...
hardware_id: ''
values: 
  - 
    key: ''
    value: ''
---

```

Should both of these messages cause an instant update, or just the first one? Does that also apply to when the level is `OK`?

From my perspective, I don't think it makes sense to update instantly on changed messages in the `OK` or `WARN` levels, but it might if the diagnostic is in the `ERROR` level. It's a pretty small change to make though, in terms of implementation - just a check on an additional field of the message.

mikepurvis:
That seems reasonable to me.

trainman419:
[...]

PR #197, opened 2021-06-04
Directly publish to /diagnostics_agg on non-ok status 

MCFurry:
I relate to [#48](https://github.com/ros/diagnostics/issues/48) and internally we started using a custom aggregator for this purpose.
We thought it could be useful to share this.

Moreover, this aggregator also fills the `/diagnostics_toplevel_state` message with the message that caused the non-ok status. We found this particularly useful to feed this to a robot operator via a display for example to have a clue about what is going on, without him needing a computer with Robot Monitor open. (Think of a display in a remote control, or a simple LCD on the robot) But also for logging purposes it could be nice to have this.

(My editor automatically removes trailing whitespaces, so sorry for those diffs..)

### DIFF
--- a/diagnostic_aggregator/CMakeLists.txt
+++ b/diagnostic_aggregator/CMakeLists.txt
@@ -38,7 +38,7 @@ target_link_libraries(diagnostic_aggregator ${Boost_LIBRARIES}
 )
 add_dependencies(${PROJECT_NAME} diagnostic_msgs_generate_messages_cpp)
 
-# Aggregator node 
+# Aggregator node
 add_executable(aggregator_node src/aggregator_node.cpp)
 target_link_libraries(aggregator_node ${catkin_LIBRARIES}
                                       ${PROJECT_NAME}
@@ -59,6 +59,7 @@ if(CATKIN_ENABLE_TESTING)
   add_rostest(test/launch/test_multiple_match.launch)
 
   add_rostest(test/launch/test_discard_stale_not_published.launch)
+  add_rostest(test/launch/test_immediate_agg_publish.launch)
 endif()
 
 catkin_install_python(
--- a/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.h
+++ b/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.h
@@ -78,11 +78,11 @@ Output:
   /Robot/Sensors/Tilt Hokuyo/Frequency
   /Robot/Sensors/Tilt Hokuyo/Connection
 \endverbatim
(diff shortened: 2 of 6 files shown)


g-gemignani:
Hi @MCFurry and thank you for your PR.
Please correct me if I am wrong, but the changes you are proposing are an ABI-breaking change. I am not sure how this kind of changes should be handled in the future... Maybe @trainman419 or @Karsten1987 can comment here since they were involved in a PR with the same issue in https://github.com/ros/diagnostics/pull/88?

MCFurry:
Thanks for your response.

I don't know about ABI compatibility, what is the main issue when breaking this?

Please advice if you agree on the functionality but have an idea about staying ABI-compatible.

g-gemignani:
The main issue with ABI compatibility is that, by modifying classes onto which other packages rely on, in order to ensure a proper functioning of the downstream packages, we would have to ask everyone to rebuild such packages again. This is why, usually, ABI breaking changes are introduced when switching versions. Since noetic is the last version that will be published for ROS1, I am not sure how we should handle such changes. That is why I was asking the opinion of people more expert than me.

MCFurry:
Ahh thanks for the explanation. I see the issue now, especially with ROS1.

I guess it will be tricky getting this feature in without ABI breaking changes then, but please advice if someone sees a way.

If the feature itself is still considered useful to others I can of course try to put this in the ROS2 version?

g-gemignani:
The feature itself is definitely useful in my opinion. What could be maybe done is to add another class that inherits and extends the class that you are modifying. A user could, in this way, use one class or another depending on if the new behavior is wanted or not. By implementing a new class and not touching the previous one, I believe that the ABI compatibility would be kept.

MCFurry:
That sounds like a neat solution indeed. If someone can advice whether this could indeed keep ABI compatibility I'll look into it!

rokusottervanger:
> That sounds like a neat solution indeed. If someone can advice whether this could indeed keep ABI compatibility I'll look into it!

I have time for this now, but I need confirmation that this direction will indeed lead to an acceptable PR.

g-gemignani:
Thank you for the PR!

Unfortunately I don't think we should merge this one.

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160).
Every change comes with a risk of introducing regressions, and there isn't much time left to fix them.
I'm closing pull requests that add features so that the remaining time is allocated towards bug fixes and compatibility with newer Ubuntu distros.

PR #198, opened 2021-06-07
Add missing lm-sensors dependency for sensors_monitor.py

vincentrou:


g-gemignani:
Hi @vincentrou and thank you for your PR. 
Why do you need to add a run_depend on lm-sensors? I might be wrong, but I do not believe this is needed.

vincentrou:
Hi @g-gemignani 
sensors_monitor.py is calling the software "sensors" here : https://github.com/ros/diagnostics/blob/noetic-devel/diagnostic_common_diagnostics/src/diagnostic_common_diagnostics/sensors_monitor.py#L172
The software "sensors" is installed by "lm-sensors".
For now, I had to install it manually on ubuntu 20.04 to make sensors_monitor.py works.

g-gemignani:
True, sorry for missing it, and thank you for the pr!

PR #200, opened 2021-08-03
Time Diagnostics can be used with Simulated Time

MarcoLm993:
This PR adds the possibility to provide an external rclcpp::Clock instance to the `TopicDiagnostic` and the `TimeStampStatus` objects at construction time.
This way the `TimeStampStatus::tick` method (and the `TopicDiagnostic::tick` one) can compare the input time to different clock sources (e.g. the one by Gazebo), depending on the rclcpp::Clock instance it was provided with.


diagnostic_updater/include/diagnostic_updater/publisher.hpp
```diff
@@ -45,6 +45,7 @@
 
 #include "rclcpp/publisher.hpp"
 #include "rclcpp/subscription.hpp"
+#include "rclcpp/clock.hpp"
```
Karsten1987, line 48: please sort alphabetically.

Karsten1987 (approved):
one nit, otherwise lgtm.

Karsten1987:
I just noticed you've targeted foxy branch. Can you please target ros2? That way we'll have it in the upcoming galactic release and I'll take care of the backporting to foxy.

MarcoLm993:
I will make a new PR targeting to ros2-devel, since this branch cannot be merged there directly.

PR #209, opened 2021-10-13
Return the actual future from async_send_request

clalancette:
Recent changes to rclcpp in Rolling make it return a structure
rather than just the future.  In order to maintain this test,
return the actual future.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

### DIFF
--- a/self_test/src/run_selftest.cpp
+++ b/self_test/src/run_selftest.cpp
@@ -91,7 +91,7 @@ public:
           }
         }
       };
-    return client_->async_send_request(request, response_received_callback);
+    return client_->async_send_request(request, response_received_callback).future;
   }
 
 private:


clalancette:
Gpr is definitely going to fail here, as this PR doesn't apply for Galactic.  We'll need to make a new galactic-devel branch off of the current `ros2-devel`, and then switch https://github.com/ros/rosdistro/blob/0cc5f30da1c12ff088239a813b862ef10a3c998e/galactic/distribution.yaml#L656-L675 over to using that.

norro (approved):
LGTM.
Will you do the merge and create the 'galactic' branch before?

clalancette:
> Will you do the merge and create the 'galactic' branch before?

I created the 'galactic' branch, and changed galactic over to using it in https://github.com/ros/rosdistro/pull/31002 .

We don't have good support for "removing" a test once it has been added to a PR, so that test is failing here.  But future PRs won't even start a `Gpr` job for PRs targeting `ros2-devel`.

With that said, thanks for the review.  Merging.

PR #211, opened 2021-10-18
Replace colcon-mixin-name in the GitHub action.

clalancette:
colcon-mixin-name no longer exists in action-ros-ci.  Instead,
use colcon-defaults with the appropriate snippet, which should
work better.  This will get rid of a warning when GitHub actions
are run.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>


norro (approved):
LGTM

PR #212, opened 2021-10-19
[ROS2][Galactic backport] Use node clock for diagnostic_aggregator and diagnostic_updater (#210)

kenji-miyake:
backport of #210 

This breaks API/ABI, but following this comment, I'd like to ask maintainers whether this can be merged.
https://github.com/ros/diagnostics/pull/210#issuecomment-946703016

@Karsten1987 Could you review this, please? :pray:

PR #215, opened 2021-10-30
[ROS2][Galactic backport] Return the actual future from async_send_request (#209)

wep21:
This is a backport PR to fix test error at https://github.com/ros/diagnostics/runs/3947786231?check_suite_focus=true in #212.
@clalancette could you review this?

### DIFF
--- a/self_test/src/run_selftest.cpp
+++ b/self_test/src/run_selftest.cpp
@@ -91,7 +91,7 @@ public:
           }
         }
       };
-    return client_->async_send_request(request, response_received_callback);
+    return client_->async_send_request(request, response_received_callback).future;
   }
 
 private:


clalancette:
I don't think this is correct for the Galactic branch; at least, as far as I remember we did this fix after Galactic.  Also, the Gpr job is failing with this in place.  And if I build diagnostics against the `galactic` branch locally, it builds fine.

So I suspect something else is wrong with the test setup for the Galactic branch, but I don't think this PR is the correct fix.

wep21:
@clalancette I tried this PR change locally and it fixes github action ci failure. https://github.com/wep21/diagnostics/pull/1.
Is the environment of ros ci different from that of ros tooling ci? Do you have any thoughts on this?

wep21:
@clalancette I found out the problem in test action in `galactic` branch and fixed it in https://github.com/ros/diagnostics/pull/216.

PR #217, opened 2021-11-09
[ROS2] Use node clock in FrequencyStatus diagnostic

MarcoLm993:
This PR adds the possibility to provide an external `rclcpp::Clock` instance to the `HeaderlessTopicDiagnostic` and the `FrequencyStatus` objects at construction time.
This way the `FrequencyStatus` diagnostic will provide accurate results even in case the clock is not running in real time (e.g. when Gazebo's Real Time Factor is not 1).


norro (approved):
LGTM

PR #218, opened 2021-11-09
[ROS2][Foxy Backport] Use node clock in FrequencyStatus diagnostic

MarcoLm993:
Backport of #217 for Foxy


norro (approved):
LGTM.
It does introduce an ABI/API change, but there is currently no quality level assigned to diagnostics that should prevent those.

PR #220, opened 2021-11-25
[ROS-NOETIC][NTPMonitor] decode bytes into string

noizetma:


g-gemignani (approved):
Thank you for the pr!

PR #221, opened 2021-12-01
[ROS2] Use regex to search AnalyzerGroup

### LINKED ISSUE #213: How can I add a new group analyzer as a plugin?
HI,
I am developing a plugin for `diagnostic_aggregator`. This will be a replacement for the existing `AnalyzerGroup`, but with different behavior.
However, my group plugin does not hierarchize correctly because `AnalyzerGroup` only recognizes `diagnostic_aggregator/AnalyzerGroup` as a group in the following line.

https://github.com/ros/diagnostics/blob/ros2-devel/diagnostic_aggregator/src/analyzer_group.cpp#L148

Since the root `AnalyzerGroup` is hard-coded `diagnostic_aggregator/AnalyzerGroup`, I have no way to avoid this.
Once the group plugins are working correctly, users will be free to create their own plugins. This would be great!
I think it will only take a few changes to fix this problem. I would be happy to help.

https://github.com/ros/diagnostics/blob/ros2-devel/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.hpp#L150

KeisukeShima:
Closes #213 

I made a modification to add the Group plugin.
This will improve the extensibility of diagnostics.

### DIFF
--- a/diagnostic_aggregator/src/analyzer_group.cpp
+++ b/diagnostic_aggregator/src/analyzer_group.cpp
@@ -42,6 +42,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <regex>
 
 PLUGINLIB_EXPORT_CLASS(diagnostic_aggregator::AnalyzerGroup, diagnostic_aggregator::Analyzer)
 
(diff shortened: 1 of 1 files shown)


KeisukeShima:
@g-gemignani Could you please take a look at this PR?

g-gemignani:
Hi @KeisukeShima, this PR is for ROS2, which unfortunately falls outside my responsibilities.

KeisukeShima:
@g-gemignani Thank you, I understand.
If you know, could you please tell me who has responsibilities for ROS2.

g-gemignani:
If I am not mistaken, @Karsten1987 is the person in charge here

KeisukeShima:
@g-gemignani Thanks!

@Karsten1987 Could you review this PR, please?

ct2034 (approved):
Thanks for your PR :+1: 
Sorry for the late reply. I am now doing the review and from my side it looks good. It is also more readable than the original code.

kenji-miyake:
@ct2034 @ralph-lange Thank you for your review! Would it be possible for you to backport this to Humble? (And probably also close #213?)

PR #222, opened 2021-12-14
Added ram monitor and unit tests

chrisflesher:
The RAM monitor is very similar to the CPU monitor, it monitors RAM percentage.


g-gemignani (approved):
Was going to suggest some style changes, but your code actually follows the style of the package. So good to go, thank you!

PR #231, opened 2022-06-23
Fix Python3 port of hd_monitor.py

Rayman:
The data returned from a socket is `bytes`. So to properly deal with
this data we'd have to concatenate as bytes and decode to string at the
end.

The old (wrong) way of concatenating results in some stray `'b'`
characters showing up in the drive names.


g-gemignani:
Thank you!

PR #233, opened 2022-07-11
Fix python3 port of ntp_monitor.py

chrisflesher:
Some fields should be strings instead of bytes


g-gemignani (approved):
Thank you!

g-gemignani (approved):
Thank you!

PR #234, opened 2022-08-01
Fix diagnostics Updater constructor NodeHandle documentation

Roboterbastler:
Previously the documentation mistakenly stated the argument `h` was used for getting the `diagnostic_period` parameter. I fixed the documentation.


g-gemignani:
Thank you!

Issue #238, opened 2022-10-28
diagnostic_updater warning: Implicit conversions on humble 


ralwing:
I'm using diagnostic_updater in humble, on which I have warnings about implicit type conversions:
```
/opt/ros/humble/include/diagnostic_updater/diagnostic_updater.hpp: In constructor ‘diagnostic_updater::Updater::Updater(std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeClockInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeLoggingInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeParametersInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeTimersInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeTopicsInterface>, double)’:
/opt/ros/humble/include/diagnostic_updater/diagnostic_updater.hpp:389:55: error: conversion from ‘double’ to ‘rcl_duration_value_t’ {aka ‘long int’} may change value [-Werror=float-conversion]
  389 |     period_(rclcpp::Duration::from_nanoseconds(period * 1e9)),
      |                                                ~~~~~~~^~~~~
/opt/ros/humble/include/diagnostic_updater/diagnostic_updater.hpp:399:57: error: conversion from ‘double’ to ‘rcl_duration_value_t’ {aka ‘long int’} may change value [-Werror=float-conversion]
  399 |     period_ = rclcpp::Duration::from_nanoseconds(period * 1e9);
      |                                                  ~~~~~~~^~~~~
/opt/ros/humble/include/diagnostic_updater/diagnostic_updater.hpp: In member function ‘void diagnostic_updater::Updater::setPeriod(double)’:
/opt/ros/humble/include/diagnostic_updater/diagnostic_updater.hpp:423:57: error: conversion from ‘double’ to ‘rcl_duration_value_t’ {aka ‘long int’} may change value [-Werror=float-conversion]
  423 |     setPeriod(rclcpp::Duration::from_nanoseconds(period * 1e9));
      |                                                  ~~~~~~~^~~~~
/opt/ros/humble/include/diagnostic_updater/diagnostic_updater.hpp: In member function ‘void diagnostic_updater::Updater::broadcast(int, std::string)’:
[...]

PR #239, opened 2022-10-28
Fix implicit conversion warnings

ralwing:
Fix for https://github.com/ros/diagnostics/issues/238


ralph-lange (approved):
LGTM. Thank you for this fix!

PR #242, opened 2022-11-16
Port ntp_monitor to ROS2

RFRIEDM-Trimble:
I decided to port the NTP monitor to ROS2. 

Out of scope: 
* Any of the other python nodes, some already exist in other pacakges
* Adding ROS params to the NTP node <- I'd like to do this in a follow up PR to minimize code changes.

Run instructions:
```
ros2 run diagnostic_common_diagnostics ntp_monitor.py  --no-self-test
```

Example diagnostics message:
```
---
header:
  stamp:
    sec: 0
    nanosec: 0
  frame_id: ''
status:
- level: "\x01"
  name: NTP offset from eb028766f08f to pool.ntp.org
  message: NTP Offset Too High
  hardware_id: eb028766f08f
  values:
  - key: Offset tolerance (us)
    value: '500'
  - key: Offset tolerance (us) for Error
    value: '5000000'
  - key: Offset (us)
    value: '56749.0'
 ```
 
 One thing unclear is the testing requirements for the python code in this repo, it seems like the python packages are not as maintained. Please advice what sort of tests we would like to pass for this branch.

PR #249, opened 2022-11-22
Port ntp_monitor to ROS2 (#242)

ct2034:
* Port ntp_monitor to ROS2
* Update to python3
* Modify cmake package structure to support ament_python_cmake
* Use argparse
* Refactor exception handling for python3
* Fix re.match to decode bytes object
* Reduce code duplication
* Handle quitting subprocess call correctly according to latest docs
* Enable linters on diagnostics_common_diagnostics
* Disable linting the files that aren't ported to ROS2
* Add an author warning for disabled files, since eventually these should be enabled
* Get lint tests to pass
* Switch depreacted author attribute to maintainer in package.xml
* use forthcoming version instead of version bump
* Parse argv rather than only sys.argv and use members
* sys args should be the 2nd and later elements, but needs to be passed in like so: https://stackoverflow.com/questions/38853812/how-to-use-python-argparse-with-args-other-than-sys-argv
* Use class member variables now that ntp_diag() has access to the variables
Signed-off-by: Ryan Friedman <ryan_friedman@trimble.com>
Co-authored-by: Christian Henkel <christian.henkel2@de.bosch.com>


ct2034:
Merge after #259

ralph-lange (approved):

Issue #254, opened 2022-11-22
Add tests for NTP Monitor


RFRIEDM-Trimble:
The NTP monitor is now ported to ROS2, but there are no tests for it.Proposed issue scope:

- [ ] Add dependency on pytest such that tests run with colcon
- [ ] Add a new python file for testing the ntp monitor
- [ ] Add a test for parsing the output of ntpdate
- [ ] Add a test for ensuring the offset threshold results in levels 0,1,2 of diagnostic status
- [ ] Any other test suggestions are welcome

 I can take assignment of this ticket and issue a PR for it.

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/what-is-the-recommended-pytest-workflow-for-an-ament-cmake-python-package/28471/1

ct2034:
see #255

PR #255, opened 2022-11-23
Test ntp monitor

RFRIEDM-Trimble:
* Refactor some functions to be more testable
* Fix a failure for parsing empty output that was trying to index a Null
value
* Use 1E6 rather than 1000000
* Add dependency on pytest
* Fix rosdep depdendency issues for existing keys (removed some duplicates, etc)
* Add tests support to run from ament_cmake_pytest
* Add missing timestamp for header from ROS2 port
* Solves https://github.com/ros/diagnostics/issues/254


ct2034:
Hi @RFRIEDM-Trimble, thanks again for your contribution.
Could you please change your pr to be towards ros2 and update your branch with it. Then please request my review

ct2034:
> Hi @RFRIEDM-Trimble, thanks again for your contribution. Could you please change your pr to be towards ros2 and update your branch with it. Then please request my review

nvm. I have the power now ;-)

ct2034:
This is no longer needed since we are not using plain `ntpdate` any more.

Ryanf55:
> This is no longer needed since we are not using plain `ntpdate` any more.

What's the replacement? Sorry, I migrated accounts.

ct2034:
It is now using a python library called `ntplib`. So it is no longer necessary to test the parsing of command line output.

PR #257, opened 2022-11-24
Fix/cleanup ros1

ct2034:
This removes all remainders of ROS1 packages that are not yet ported to ROS2. If these are required for later porting efforts, it is of cause possible to get them from the ROS1 branches.


ralph-lange:
LGTM. Thank you!

PR #267, opened 2022-12-16
Fixing porting remainders

### LINKED ISSUE #262: Humble diagnostic_common_diagnostics build error
For some reason `package.xml` of this package has `catkin` `buildtool` tag that seems off for a ROS 2 package. It produces error such as 

```
The manifest of package "diagnostic_common_diagnostics" (with format version 1) must not contain the following tags: exec_depend.
Either update to a newer format or replace <exec_depend> tags with <run_depend> tags.
```

I think the correct way is just like `foxy` and `galactic` branch meaning the `buildtool` tag must be `ament_cmake` and `ament_cmake_python`. A pull request #261 have been opened by another user as well that fixes it.

ct2034:
Fixes #262


ralph-lange (approved):
LGTM.

PR #269, opened 2022-12-23
Self test publishes the service under the node name, again

ct2034:
I am not sure if this is supposed to work. But the original ROS1 wiki indicated that there will be service called `self_test` per node. I can not see how the code will call them. If I run the code as it was on rolling, there will only be one self_test service that will be called. Advertising this same service from multiple nodes obviously doesn't work.


ct2034:
To document it here: I found out that the current ros2 implementation differs from the one in ros1. In particular, I compared `noetic-devel` @ [12ba2ee](https://github.com/ros/diagnostics/commit/12ba2ee06d2609286cd2f38a22f3a176dd051da0) to `rolling` @ [12ab006](https://github.com/ros/diagnostics/commit/12ab0065178a2ebd6b205232af2dea8dff6b8cde).

ROS1 did create a service under `[NODE_NAME]/self_test` and you could perform the self_test with `run_selftest [NODE_NAME]`.
On ROS2, the advertised service is not under the node's namespace, but simply called `self_test` and also the `run_selftest` command does not take an argument. This works as long as there is only one node that advertises this service, but *not with multiple nodes that implement a self test*.

ct2034:
Hi @trainman419 @bricerebsamen @Karsten1987 
Could you please comment on this, why the API did change like that for ROS2 in the first place?
This change tries to recreate the API as it was in ROS1, because I think it makes more sense like this.

ct2034:
Friendly ping  @trainman419 @bricerebsamen @Karsten1987

self_test/include/self_test/test_runner.hpp
```diff
@@ -155,15 +160,13 @@ class TestRunner : public diagnostic_updater::DiagnosticTaskVector
       };
 
     service_server_ = rclcpp::create_service<diagnostic_msgs::srv::SelfTest>(
-      base_interface_, service_interface_, "self_test", serviceCB, rmw_qos_profile_default,
-      nullptr);
+      base_interface_, service_interface_,
+      std::string(base_interface_->get_name()) + std::string("/self_test"),
```
ct2034, line 164: most important change here ...

PR #271, opened 2022-12-26
use hpp instead of deprecated .h pluginlib headers

lucasw:


lucasw:
whoops did a search and replace on the readme

PR #272, opened 2023-01-20
Rolling branch builds on foxy and humble

ct2034:
With one switch, the rolling code compiles for foxy and humble

TODO:
- [x] Compare with foxy
- [x] Compare with humble
- [ ] setup release accordingly


ralph-lange (approved):
LGTM

PR #279, opened 2023-01-30
Fix/ros2 no rclpy in cmakelists

ct2034:
This should avoid errors like https://build.ros2.org/job/Rbin_uJ64__diagnostic_common_diagnostics__ubuntu_jammy_amd64__binary/4/console#console-section-1


ralph-lange (approved):
Thank you, @ct2034.

PR #282, opened 2023-02-04
Fix two code-analyser bugs

Timple:
We have a static code analyser running and it triggered on these two statements.

The `return 0` can never be reached. 

And one should only invert booleans with `!`. Not integers. The `std::set` used has an `.insert` option which returns if it was actually added. So using that instead.


peci1 (approved):
As this has been successfully merged into ro2, I don't see any reason to not merge it to noetic-devel, too.

PR #283, opened 2023-02-04
Fix code-analyser bug

Timple:
Same as #282 but for rolling


ralph-lange (approved):
LGTM. Thank you for this improvement! But could you please raise this PR to the new ros2 branch, from which Foxy, Humble, and Rolling have been released together just a few days ago? (I will rename the branches foxy, humble, and rolling soon to indicate that they are stale and will also change the default branch.)

Timple:
Done

ralph-lange (approved):
LGTM. Thank you for the improvement!

PR #285, opened 2023-02-10
Use default implementation for copy constructor

jordan-palacios:
Copy constructor definition is missing. I added the default one.


ct2034 (approved):
Thanks @jordan-palacios for your contribution :grin:

PR #290, opened 2023-02-15
Deleting copy constructor

ct2034:
Judging by this comment
```
   * Defined and marked explicit so that you don't accidentally use a
   * function<void(DiagnosticStatusWrapper)> where a function<void(DiagnosticStatusWrapper &)>
   * is needed. Otherwise, it's easy to accidentally create a DiagnosticTask that silently does
   * nothing.
``` 
from https://github.com/ct2034/diagnostics/blob/2707b64905b1312783b8c91caaa16418a7ad0376/diagnostic_updater/include/diagnostic_updater/diagnostic_status_wrapper.hpp#L76, it is a good idea **not** to implement the copy constructor here. 
It was removed by https://github.com/ros/diagnostics/pull/285, do you @jordan-palacios have a concrete use case where you need a copy constructor?

### DIFF
--- a/diagnostic_updater/include/diagnostic_updater/diagnostic_status_wrapper.hpp
+++ b/diagnostic_updater/include/diagnostic_updater/diagnostic_status_wrapper.hpp
@@ -79,7 +79,7 @@ public:
    * nothing.
    * \param other Reference to object to copy
    */
-  explicit DiagnosticStatusWrapper(const DiagnosticStatusWrapper & other) = default;
+  explicit DiagnosticStatusWrapper(const DiagnosticStatusWrapper & other) = delete;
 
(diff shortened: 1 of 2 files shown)


jordan-palacios:
Hello @ct2034, not really.

I came upon the issue while porting some code from ROS1 to ROS2. I ended up not needing the copy constructor at all. Keep in mind though, the copy constructor it's part of the ROS1 implementation so maybe I won't be the last one to trip on this issue.

ct2034:
Okay, thanks for your comment. I suggest to going through with this, then. I mean, so far it can not have been used in ros2, because it was not implemented. If someone needs it in the future, I am happy to look into this again.

PR #299, opened 2023-04-25
Replace Clock().now() to node.get_clock().now() in diagnostic_updater.py

h-wata:
### Description
In this PR, the `Clock` import in the diagnostic_updater.py has been replaced with `node.get_clock()`. 
This ensures that the correct `ROS_TIME` is used when the `use_sim_time` parameter is set in the `ros2 node`.

### Environment
ros 2 : Humble

### Ref

* https://answers.ros.org/question/321536/replacement-for-rospytimenow-in-ros2/
* https://design.ros2.org/articles/clock_and_time.html

### Verifying the Changes

* Terminal1 

```bash
ros2 launch gazebo_ros gazebo.launch.py
```

* Terminal2

```bash
ros2 run diagnostic_updater example.py --ros-args -p use_sim_time:=true # or false
```

* Terminal3

```bash
ros2 topic echo /diagnostics --field header.stamp.sec
```

#### Before
  header.stamp is `System Time` regardless of `use_sim_time`.

#### After
   header.stamp is `ROS Time` when `use_sim_time` is true.

### DIFF
--- a/diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
+++ b/diagnostic_updater/diagnostic_updater/_diagnostic_updater.py
@@ -42,7 +42,6 @@ import threading
 
 from diagnostic_msgs.msg import DiagnosticArray
 from diagnostic_msgs.msg import DiagnosticStatus
-from rclpy.clock import Clock
 
 from ._diagnostic_status_wrapper import DiagnosticStatusWrapper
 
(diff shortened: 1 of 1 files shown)


ct2034:
Thanks for your pull request 👍

ct2034 (approved):

PR #305, opened 2023-05-24
Empty default aggregator base_path

### LINKED ISSUE #303: [ros2] Diagnostics aggregator base path
Hi,

With a diagnostic named `led1`:
```
analyzers:
  ros__parameters:
    path: Aggregator
    hardware:
      type: 'diagnostic_aggregator/GenericAnalyzer'
      path: Hardware
      startswith: [ 'led' ]
```

The name for that diagnostic  in the topic `/diagnostics_agg`  would be `/Aggregator/Hardware/led1` 

However, if I don't set the base path:
```
analyzers:
  ros__parameters:
    hardware:
      type: 'diagnostic_aggregator/GenericAnalyzer'
      path: Hardware
      startswith: [ 'led' ]
```
The name would be `//Hardware/led1`.
With the same case in ROS, the name would be `/Hardware/led1`

Is there any reason why the default `base_path` for the Diagnostics Aggregator in ROS 2 is `/`?

ct2034:
Fixes #303

### DIFF
--- a/diagnostic_aggregator/CMakeLists.txt
+++ b/diagnostic_aggregator/CMakeLists.txt
@@ -81,7 +81,8 @@ if(BUILD_TESTING)
   set(create_analyzers_tests
     "primitive_analyzers"
     "all_analyzers"
-    "analyzer_group")
+    "analyzer_group"
+    "empty_root_path")
 
   foreach(test_name ${create_analyzers_tests})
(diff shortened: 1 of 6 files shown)


Noel215:
Since the base path now is empty, this line must be also changed in order to have the initial `/` when having a `base_path`
https://github.com/ros/diagnostics/blob/369a769d2e8fd1babe97526f45d15628a76c0b0d/diagnostic_aggregator/src/aggregator.cpp#L82
to
```
      base_path_ = "/" + param.second.as_string();
```

PR #315, opened 2023-09-21
Avoid rolling up an ERROR state when empty GenericAnalyzer blocks are marked discard_stale, or when all of their items are STALE.

asymingt:
Add logic to a parent `AnalyzerGroup` to prevent rolling up an `ERROR` state when a `GenericAnalyzer` child block is marked with `discard_stale: true` and either of these conditions are met:

1. There are no statuses that match any of the `GenericAnalyzer` child blocks (its size is zero).
2. Every matching status in the `GenericAnalyzer` child block has been marked stale.
 
In this example, if `bar` and `baz` have no matching statuses or all of their statuses are `STALE`, they will roll up as `OK` because the `discard_stale: true` flag implies that stale statuses are disposable.

``` 
diagnostics_aggregator:
  ros__parameters:
    pub_rate: 1.0
    path: 'robot'
    analyzers:
      part:
        type: 'diagnostic_aggregator/AnalyzerGroup'
        path: 'part'
        foo:
          type: 'diagnostic_aggregator/GenericAnalyzer'
          path: 'foo'
          find_and_remove_prefix: ['/foo:']
          num_items: 1
        bar:
          type: 'diagnostic_aggregator/GenericAnalyzer'
          path: 'bar'
          find_and_remove_prefix: ['/bar:']
          discard_stale: true
        baz:
          type: 'diagnostic_aggregator/GenericAnalyzer'
          path: 'baz'
          find_and_remove_prefix: ['/baz:']
          discard_stale: true
```

### DIFF
--- a/diagnostic_aggregator/test/test_discard_behavior.py
+++ b/diagnostic_aggregator/test/test_discard_behavior.py
@@ -117,7 +117,7 @@ TEST_METADATA = [
         foo_status=None,
         bar_discard=False,
         bar_status=None,
-        agg_expected=DiagnosticStatus.ERROR, # <-- This is the case we are testing for.
+        agg_expected=DiagnosticStatus.ERROR,  # <-- This is the case we are testing for.
         # if one of the children is *not* marked discard_stale := true and
         # there are no statuses, then the parent should roll up to ERROR.
     ),
@@ -133,7 +133,7 @@ TEST_METADATA = [
         foo_status=None,
         bar_discard=False,
         bar_status=DiagnosticStatus.OK,
-        agg_expected=DiagnosticStatus.OK, # <-- This is the case we are testing for.
+        agg_expected=DiagnosticStatus.OK,  # <-- This is the case we are testing for.
         # but if a child is marked discard_stale := true and there are no statuses,
         # the parent should roll up to OK.
     ),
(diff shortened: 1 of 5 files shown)


ct2034:
@asymingt - thanks a lot for your contribution!
Please also add a test case that checks for this and that defines the desired behavior.

asymingt:
> @asymingt - thanks a lot for your contribution! Please also add a test case that checks for this and that defines the desired behavior.

I have done my best to add one, given the limited time available to me!

diagnostic_aggregator/test/test_discard_behavior.py
```diff
@@ -0,0 +1,295 @@
...
+        bar_discard=False,
+        bar_status=None,
+        baz_result=DiagnosticStatus.ERROR,
+    ),
+    TestMetadata(
+        foo_discard=True,
+        foo_status=None,
+        bar_discard=False,
+        bar_status=DiagnosticStatus.OK,
+        baz_result=DiagnosticStatus.ERROR,
```
asymingt, line 134: This case doesn't please me...

ct2034:
@asymingt I think I am happy with this now. But could you please have look at the comments and changes I made in 6e852dc73f4a93efc2e60e419969192d9bf22f4f and just confirm if they make sende to you?

ct2034 (approved):

asymingt:
> @asymingt I think I am happy with this now. But could you please have look at the comments and changes I made in [6e852dc](https://github.com/ros/diagnostics/commit/6e852dc73f4a93efc2e60e419969192d9bf22f4f) and just confirm if they make sende to you?

Yup, those changes make sense. LGTM. Thanks, @ct2034.

Issue #318, opened 2023-11-02
Aggregator and updater with fully qualified node name


outrider-jhulas:
With the current implementation in ROS2 it's impossible to aggregate the diagnostic messages based on the full node names, if the diagnostic message has been created using `diagnostic_updater`. 
  
In many cases, the node name is enough to fully identify the source of the diagnostic message.
For example we have a sensor driver node with the name `my_sensor`. In many real application there will be more than one instance of a node and the name will be remapped and namespaces will be added. 
So The FQN will be:

`/robot/sensor/rear_left`

In this case the aggregator can only sort the diagnostic messages based on the name which is `rear_left`. What if we also had a `robot/wheel/rear_left`?

Even the [example](https://github.com/ros/diagnostics/tree/ros2/diagnostic_aggregator#example) in `diagnostic_aggregator` would be impossible to achieve if the diagnostic messages were published using the `diagnostic_updater`.

One solution would be to get the fully qualified name in the [updater]. (https://github.com/ros/diagnostics/blob/24fb79aeb577f2e2c089021406e73cafd3c86ea9/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp#L394)

PR #320, opened 2023-11-23
feat: add param to use fqn in updater

outrider-jhulas:
As mentioned in #318, it's a possible solution to the fully qualified name not being included anywhere in the diagnostic message, making it impossible to properly aggregate. 

This PR adds a parameter `use_fqn`  to set a bool. It will then use fully qualified name in the `name` field in the diagnostic updater


ct2034:
Thanks for your contribution @outrider-jhulas 😀 
Could you please add a test that verifies what names are used with and without your parameter.

PR #324, opened 2023-12-21
Aggregator: publish diagnostics_toplevel_state immediately on every degradation

Timple:
Enhancement to: https://github.com/ros/diagnostics/pull/317

Notes
- This change is backward compatible with the state before 2023-12-05.
- This state is **not** backward compatible with the changes since then.
I chose this option as there has not been a release since then. And yet another backward compatible parameter seemed overdone. 

@outrider-jhulas does this work for your situation as well?

@outrider-jhulas I could not get your test to work. `colcon test` always succeeds but `launch_test diagnostic_aggregator/test/test_critical_pub.py` always fails.

Rationale: we respond quickly to warnings as well as errors. So for consistency all degradation should be published quickly i.m.o.

### DIFF
--- a/diagnostic_aggregator/src/aggregator.cpp
+++ b/diagnostic_aggregator/src/aggregator.cpp
@@ -156,14 +156,13 @@ void Aggregator::diagCallback(const DiagnosticArray::SharedPtr diag_msg)
     for (auto j = 0u; j < diag_msg->status.size(); ++j) {
       analyzed = false;
 
-      const bool top_level_state_transition_to_error =
-        (last_top_level_state_ != DiagnosticStatus::ERROR) &&
-        (diag_msg->status[j].level == DiagnosticStatus::ERROR);
+      const bool top_level_state_degradation = (last_top_level_state_ < diag_msg->status[j].level);
 
-      if (critical_ && top_level_state_transition_to_error) {
+      if (critical_ && top_level_state_degradation) {
         RCLCPP_DEBUG(
-          logger_, "Received error message: %s, publishing error immediately",
+          logger_, "Received degrated message: %s, publishing immediately",
           diag_msg->status[j].name.c_str());
+
         DiagnosticStatus diag_toplevel_state;
         diag_toplevel_state.name = "toplevel_state_critical";
         diag_toplevel_state.level = diag_msg->status[j].level;
--- a/diagnostic_aggregator/src/aggregator.cpp
+++ b/diagnostic_aggregator/src/aggregator.cpp
@@ -151,27 +151,11 @@ void Aggregator::diagCallback(const DiagnosticArray::SharedPtr diag_msg)
   checkTimestamp(diag_msg);
 
   bool analyzed = false;
+  bool immediate_report = false;
   {  // lock the whole loop to ensure nothing in the analyzer group changes during it.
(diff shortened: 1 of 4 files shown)


Timple:
I took the liberty of mimicking this implementation: https://github.com/ros/diagnostics/pull/197, as that has proven itself in production for a few years now. 
As a bonus, it re-uses the `publishData` function such that the actual diagnostics are available for analysis.

Timple:
Happy new year everyone!

As this PR is not backwards compatible with changes as of 2023-12-05, I guess it might be preferred to get this in quickly?

outrider-jhulas:
Hi, Sorry for the slow answer, your MR looks good to me. 

I couldn't really replicate the test issue

ct2034 (approved):
Thanks for your contribution

Timple:
You're welcome!

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -182,8 +165,17 @@ void Aggregator::diagCallback(const DiagnosticArray::SharedPtr diag_msg)
       if (!analyzed) {
         other_analyzer_->analyze(item);
       }
+
+      // In case there is a degraded state, publish immediately
+      if (critical_ && item->getLevel() > last_top_level_state_) {
```
ct2034, line 170: This could lead to problems, because `stale` is larger than everything else.

ct2034 (changes requested):
Sorry, I have to correct my previous review. 
- Please see inline comment
- Please add a test for this behavior in https://github.com/ros/diagnostics/blob/add/cpu_monitor/diagnostic_aggregator/test/test_critical_pub.py#L89-L90

diagnostic_aggregator/src/aggregator.cpp
Timple: Yes? Would it be bad to publish immediately if items become stale?

Timple: @ct2034 is this a problem if we publish immediately on stale?

Timple:
I've added the test, ready for a re-review

Timple:
Subtle ping :slightly_smiling_face:

Timple:
Since this is technically a breaking change, I would like to have this in rolling before Jade is split off.

@ct2034 are you happy with the added test?

Timple:
Aaaand we missed Jazzy :crying_cat_face: 

Can we please get some traction on this now? Or have at least some comments on what needs to be changed?

Timple:
Monthly ping?

diagnostic_aggregator/src/aggregator.cpp
ct2034: No, I guess you are right

ct2034 (approved):

ct2034:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|ros2-humble|[](https://github.com/ros/diagnostics/pull/355)|
|✅|ros2-iron|[](https://github.com/ros/diagnostics/pull/356)|
|✅|ros2-jazzy|[](https://github.com/ros/diagnostics/pull/357)|

### Questions ?

Timple:
Great, thank you!

PR #329, opened 2024-01-22
Add add_analyzer functionality

MartinCornelis2:
Applied the following changes to allow adding analyzers at runtime similar to the ROS1 functionality. 

- Move the creation of `analyzer_group_ ` and `other_analyzer_` from the constructor to a new function `initAnalyzers()`.
- Added a subscriber to parameter events `param_sub_` that triggers a callback `parameterCallback`.
- Check in the callback if `this` node got any new paramters -> if true, lock the mutex and call `initAnalyzers()` again.

- Created an `add_analyzer` node that forwards its own parameters to the `diagnostic_aggregator` by sending a request to `/diagnostics_agg/set_parameters_atomically` service

- **(EDIT) [The `add_analyzer` looks for all parameters starting with the prefix `analyzers.`,] has been replaced with [The `add_analyzer` looks for all parameters for which a similar named parameter exists with the suffix ".path"]** converts them to a parameter_msg and sends them together as one `SetParametersAtomically::Request`

Example usage:
```
<node name="add_my_analyzer" pkg="diagnostic_aggregator" exec="add_analyzer" >
        <param from="$(find-pkg-share <pkg_name>)/param/<param_file_path>.yaml" allow_substs="true"/>
        <remap from="/diagnostics_agg/set_parameters_atomically" to="/diagnostics_autonomy/set_parameters_atomically" />
</node>
```

### DIFF
--- a/diagnostic_aggregator/CMakeLists.txt
+++ b/diagnostic_aggregator/CMakeLists.txt
@@ -162,6 +162,7 @@ ament_python_install_package(${PROJECT_NAME})
 
 # Install Example
 set(ANALYZER_PARAMS_FILEPATH "${CMAKE_INSTALL_PREFIX}/share/${PROJECT_NAME}/example_analyzers.yaml")
+set(ADD_ANALYZER_PARAMS_FILEPATH "${CMAKE_INSTALL_PREFIX}/share/${PROJECT_NAME}/example_add_analyzers.yaml")
 configure_file(example/example.launch.py.in example.launch.py @ONLY)
 install( # launch descriptor
   FILES ${CMAKE_CURRENT_BINARY_DIR}/example.launch.py
@@ -172,7 +173,7 @@ install( # example publisher
   DESTINATION lib/${PROJECT_NAME}
 )
 install( # example aggregator configration
-  FILES example/example_analyzers.yaml
+  FILES example/example_analyzers.yaml example/example_add_analyzers.yaml
   DESTINATION share/${PROJECT_NAME}
 )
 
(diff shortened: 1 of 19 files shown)


ct2034 (changes requested):
Thanks for adding this feature.
Please add 
- information on the usage in diagnostic_aggregator/README.md
- a test that initializes a new analyzer at runtime

diagnostic_aggregator/CMakeLists.txt
```diff
@@ -129,6 +130,27 @@ if(BUILD_TESTING)
     )
   endforeach()
 
+  set(add_analyzers_tests
+  "all_analyzers")
+
+  foreach(test_name ${add_analyzers_tests})
```
MartinCornelis2, line 136: I followed the previous pattern even though I only provide a test for one of them. I can remove the loop or add the expected outcome for the other analyzers, or just leave this as is.

Timple:
Looks like all comments where addressed. Any chance for a re-review?

diagnostic_aggregator/CMakeLists.txt
ct2034: thats ok. This way it could be extended later

Timple:
If I may be so bold... Since there's a lot of maintenance effort going on: can this get a re-review? :slightly_smiling_face:

ct2034:
Of course. Let's use the momentum while it lasts ;-)

ct2034 (approved):

ct2034:
https://build.ros2.org/job/Rpr__diagnostics__ubuntu_noble_amd64/27/ fails due to https://github.com/PyCQA/flake8/issues/646

ct2034:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|ros2-humble|[](https://github.com/ros/diagnostics/pull/359)|
|✅|ros2-iron|[](https://github.com/ros/diagnostics/pull/360)|
|✅|ros2-jazzy|[](https://github.com/ros/diagnostics/pull/361)|

### Questions ?

PR #331, opened 2024-02-20
Minimize header includes by moving impl to .cpp files

Rayman:
I'm trying to improve compilation times. One obvious candidate was this package. I've tried to remove includes of rclcpp (which is huge) and move all implementations to a cpp file. I did not modify any code, only move it.

### DIFF
--- a/diagnostic_updater/src/example.cpp
+++ b/diagnostic_updater/src/example.cpp
@@ -34,8 +34,9 @@
 
 #include <diagnostic_updater/diagnostic_updater.hpp>
 #include <diagnostic_updater/publisher.hpp>
-
 #include <std_msgs/msg/bool.hpp>
+
+#include "rclcpp/rclcpp.hpp"
 
 using namespace std::chrono_literals;
 
--- a/diagnostic_updater/CMakeLists.txt
+++ b/diagnostic_updater/CMakeLists.txt
@@ -17,16 +17,18 @@ find_package(rclcpp REQUIRED)
 find_package(rclpy REQUIRED)
 find_package(std_msgs REQUIRED)
 
-add_library(${PROJECT_NAME} INTERFACE)
+add_library(${PROJECT_NAME}
+  src/diagnostic_updater.cpp
+)
 target_include_directories(
   ${PROJECT_NAME}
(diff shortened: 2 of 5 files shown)


Rayman:
Hello, can I get some feedback on my PR?

ct2034:
Hi @Rayman. To be honest, I prefer the header-based approach right now for simplicity. Do you know that this has a noticeable speed improvement?

Rayman:
The templates from rclcpp are relatively heavy, especially `create_subscriber`. So by moving them to the implementation files we can avoid recompiling them. `diagnostics` only uses `create_publisher`, which is less heavy but still the same argument applies.

Also the include `rclcpp/rclcpp.hpp` includes everything, and is pretty huge. Most of the time in the frontend is spent parsing these headers.

With this PR our software still compiles perfectly fine, so no change needed in dependent software

If you want I can show you some measurements, the results are not big but multiplied by every node in our system (100s) it becomes noticeable.

diagnostic_updater/src/example.cpp
```diff
@@ -34,9 +34,10 @@
 
 #include <diagnostic_updater/diagnostic_updater.hpp>
 #include <diagnostic_updater/publisher.hpp>
-
 #include <std_msgs/msg/bool.hpp>
 
+#include "rclcpp/rclcpp.hpp"
+
```
ct2034, line 40: I see that it was probably wrong that this was missing before. But now it is technically a breaking change.

Rayman: Yes it is a breaking change. So this should not be merged into iron, but in rolling. I'll then get released for jazzy

Rayman:
I did some quick tests with a simple node that has a diagnostic_updater. The compilation time went from 5 to 2.5 seconds. I only measured this once so there could be some variability.

compile-master
compile-branch

ct2034 (approved):
Thanks for you contribution

ct2034:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|ros2|[](https://github.com/ros/diagnostics/pull/389)|

### Questions ?

PR #336, opened 2024-03-22
Fix release target documentation

Rayman:
Is see that there are now two branches `ros2` and `ros2-jazzy`. The second branch should target `jazzy` and newer distros, including `rolling`.

To make this more clear you could also change the branch names:
`ros2` -> `ros2-iron` (iron and older)
`ros-jazzy` `ros2` (jazzy and newer)

Or make 1 branch per release
`ros2-humble`
`ros2-iron`
`ros2`


ct2034:
Yes, you are absolutely right. Sorry, I made the changes already

PR #341, opened 2024-04-26
Add timestamp to diagnotics msg

tonynajjar:
Some visualization tools require the timestamp


tonynajjar:
side-question: any reason there are two timers, one for pubCB and one for checkCB? Can't we just combine them and make the frequency configurable? I can do it within this same PR if you agree

tonynajjar:
superseeded by https://github.com/ros/diagnostics/pull/342

PR #354, opened 2024-06-17
Add starting_up_state parameter to Updater

redvinaa:
In the current implementation, when the Updater object is constructed, it sends off an "OK" signal. This could be problematic if we consider "Everything is *running* as expected" (definition of "OK" status as per REP107), that running is already initialized.

The default behavior doesn't change with this PR, so it wouldn't break anything.
Also, in the python implementation, None can be passed to skip "Node starting up" status publishing altogether.


redvinaa:
Hi! Can I get an update on this one?

If it wasn't merged because of this test case failing, can I get some help understanding the error? I'm sure a few minutes of someone who already understands these template-based tests can save me hours of trial and error.
After that I can also rebase to fix the conflict.

ct2034:
Hi @redvinaa.
Could we please have a quick discussion about this? I am not sure if it is worth the extra overhead. Yes, 
> "Everything is running as expected" (definition of "OK" status as per REP107)

On the other hand, I also would not like my system starting in an error state. Your issue could also be solved by additional state, such as proposed in https://github.com/ros2/common_interfaces/issues/268. What is your opinion about that?

redvinaa:
Hi! Sorry for the long delay. It's possible that it would make sense to add new states, but that's a huge change and it appears to be moving very slowly. This PR on the other hand is not breaking, and I think is still useful until the new states get added.

ct2034 (approved):

diagnostic_updater/src/diagnostic_updater.cpp
```diff
@@ -190,9 +193,13 @@ void Updater::publish(std::vector<diagnostic_msgs::msg::DiagnosticStatus> & stat
 
 void Updater::addedTaskCallback(DiagnosticTaskInternal & task)
 {
+  if (starting_up_status_ < 0) {
+    return;
+  }
```
ct2034: what is the purpose of this line?

ct2034 (changes requested):

diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp
```diff
@@ -363,19 +363,23 @@ class Updater : public DiagnosticTaskVector
...
    * \param period Value in seconds to set the update period
+   * \param starting_up_status Diagnostic status to send as first message
    * \note The given period value not being used if the `diagnostic_updater.period`
    * ros2 parameter was set previously.
    */
   template<class NodeT>
-  explicit Updater(NodeT node, double period = 1.0)
+  explicit Updater(
+    NodeT node, double period = 1.0,
+    int starting_up_status = diagnostic_msgs::msg::DiagnosticStatus::OK)
```
ct2034:
```suggestion
    unsigned char starting_up_status = diagnostic_msgs::msg::DiagnosticStatus::OK)
```

diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp
```diff
@@ -385,7 +389,8 @@ class Updater : public DiagnosticTaskVector
     std::shared_ptr<rclcpp::node_interfaces::NodeParametersInterface> parameters_interface,
     std::shared_ptr<rclcpp::node_interfaces::NodeTimersInterface> timers_interface,
     std::shared_ptr<rclcpp::node_interfaces::NodeTopicsInterface> topics_interface,
-    double period = 1.0);
+    double period = 1.0,
+    int starting_up_status = diagnostic_msgs::msg::DiagnosticStatus::OK);
```
ct2034:
```suggestion
    unsigned char starting_up_status = diagnostic_msgs::msg::DiagnosticStatus::OK);
```

diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp
```diff
@@ -478,6 +483,7 @@ class Updater : public DiagnosticTaskVector
   rclcpp::Publisher<diagnostic_msgs::msg::DiagnosticArray>::SharedPtr publisher_;
   rclcpp::Logger logger_;
 
+  int starting_up_status_;
```
ct2034:
```suggestion
  unsigned char starting_up_status_;
```

diagnostic_updater/src/diagnostic_updater.cpp
```diff
@@ -44,7 +44,9 @@ Updater::Updater(
   std::shared_ptr<rclcpp::node_interfaces::NodeLoggingInterface> logging_interface,
   std::shared_ptr<rclcpp::node_interfaces::NodeParametersInterface> parameters_interface,
   std::shared_ptr<rclcpp::node_interfaces::NodeTimersInterface> timers_interface,
-  std::shared_ptr<rclcpp::node_interfaces::NodeTopicsInterface> topics_interface, double period)
+  std::shared_ptr<rclcpp::node_interfaces::NodeTopicsInterface> topics_interface,
+  double period,
+  int starting_up_status)
```
ct2034:
```suggestion
  unsigned char starting_up_status)
```

ct2034 (changes requested):

diagnostic_updater/src/diagnostic_updater.cpp
redvinaa: I don't remember anymore, removing

redvinaa:
I think the xmllint failure is not coming from this PR (no xml was modified). Otherwise, all requested changes have been addressed.

PR #364, opened 2024-06-27
Stop flooding the terminal by default with all matches, switch to debug if desired

MCFurry:
We noticed that when extending a ROS system with many analyzers, the terminal during startup is flooded with all the analyzer matches.
I reckon these can be debug logs that one can enable when desired.

(Example in the tests, since these require these stdout logs)


Timple:
```
The following tests FAILED:
	  2 - ntp_monitor_launchtest (Failed)
```
This monitor was not modified in this PR. I'll rebase to see if that helps.

Timple:
Ping @ct2034 Any objections here?

Timple:
Or can this at least get in before Kilted? :slightly_smiling_face: 
It's coming up! https://discourse.ros.org/t/upcoming-ros-2-feature-freeze-for-kilted-kaiju-on-april-14th/43096/2

diagnostic_aggregator/include/diagnostic_aggregator/generic_analyzer_base.hpp
```diff
@@ -119,7 +119,7 @@ class GenericAnalyzerBase : public Analyzer
 
     has_initialized_ = true;
 
-    RCLCPP_INFO(
+    RCLCPP_DEBUG(
```
ct2034: Can we please keep those messages that are not triggered per match on info

ct2034 (changes requested):

ct2034:
Hey @MCFurry @Timple 
Thanks for the contribution.
I am a little reluctant against changing the logging level in the code, because that is something that can be done depending on the runtime.
But I agree that it is generally a good idea to not have messages that are emitted repeatedly (like per match) in info.
So, I am asking you to please change all those messages that are emitted only once per initialization back to info.

Timple:
Sorry it took a while!

ct2034:
Based on what we discussed above, that I am happy to change the level in the `match` method. I would therefore expect no changes in the `init` method of `diagnostic_aggregator/src/analyzer_group.cpp`. Or are there any reasons to also change those tha I am missing right now?

In any case, I think this would also be a good use for [RCLCPP_DEBUG_THROTTLE](https://docs.ros2.org/latest/api/rclcpp/logging_8hpp.html#a11ad0ece390eff7363cc60c0cd55a3ba). What do you think?

Timple:
The issue is that there a loop around all parameters:
```c++
for (const auto & param : parameters) {
```

Every analyzer added prints information about all other analyzers as well. Since we have a lot of analyzers, adding only 1 causes a massive print flood.

Timple:
@ct2034 if you like things to be verbose, do you have any ideas how to get around printing always all parameters?

diagnostic_aggregator/src/analyzer_group.cpp
```diff
@@ -175,7 +175,7 @@ bool AnalyzerGroup::init(
     init_ok = false;
     RCLCPP_ERROR(logger_, "No analyzers initialized in AnalyzerGroup '%s'", n->get_namespace());
   } else {
-    RCLCPP_INFO(
```
ct2034, line 178: Could we not change these? They don't get executed per match, or do they?

diagnostic_aggregator/src/analyzer_group.cpp
```diff
@@ -238,7 +238,7 @@ bool AnalyzerGroup::match(const std::string & name)
     match_name = mtch || match_name;
     matched_[name].at(i) = mtch;
     if (mtch) {
-      RCLCPP_INFO(
```
ct2034, line 241: Could we not change these? They don't get executed per match, or do they?

ct2034 (changes requested):

diagnostic_aggregator/src/analyzer_group.cpp
Timple: Not per match, but per init. And every time a`add_analyzer` triggers a parameter event, it fully destroys and fully rebuilds the entire `AnalyzerGroup`. So this does lead to excessive logging.

ct2034: Okay, so I guess the flooding here only ocurrs in scenarios with dynamic configuration.
Could we do something where it is an info if the aggregator is new and else debug? Would this work. It would make the implementation slightly more complex, though

Timple: In anticipation of this request I already dived into that.

- Pragmatic approach:
  - Track existing analyzer names before re-init, then suppress logs for already-known ones.

- Support incremental adds
  - Quite significant rewrite (including some bookkeeping).

- Making log level a user preference
  - Quite easy with a parameter I guess.
  
 I am currently trying option 1

Timple: https://github.com/ros/diagnostics/commit/1c8d8b84504c1b4fc4b34249a7f5ed7e72d3d210 

I guess this is a bit much for only a log statement..

Timple: Perhaps a compromise: We'll make this a `ROS_INFO` but the matches (which are much more verbose) a `ROS_DEBUG`?

ct2034: Yes, I think that is a good compromise :)

Timple: Done!

ct2034 (approved):

PR #399, opened 2024-09-02
diagnostic_updater: Added documentation of the Python interface.

peci1:
I've added a Sphinx config to generate Python docs for the diagnostic_updater package.

To retain references to the C++ API, I haven't generated the standard disambiguation page with C++/Python/Msg links, but I've kept C++ API docs in the original path, generated Python into a subfolder, and added a link to the Python API docs to the C++ docs.

The behavior of the code should not change, although I've added the `__all__` variable to `__init__.py` to convince Sphinx to generate the package-level docs.

### DIFF
--- a/diagnostic_updater/.gitignore
+++ b/diagnostic_updater/.gitignore
@@ -1,0 +1,2 @@
+doc/html
+doc/manifest.yaml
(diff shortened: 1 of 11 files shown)


g-gemignani (approved):
Thank you for your contribution!

PR #418, opened 2024-12-12
feat: add `publish_values` parameter

MrBlenny:
When operating under bandwidth-constrained conditions it useful to have the option to turn off the `values` in the DiagnosticStatus message. These contain lots of redundant information in the parent diagnostics.


ct2034 (changes requested):
Thanks for your contribution. Can you please add a better explanation of that the published values are.

ct2034 (approved):

ct2034:
merged via #597

Issue #321, opened 2023-12-05
Make diagnostic aggregator composable


jlack1987:
Should be a relatively easy feature to add to make the diagnostics aggregator composable so it can be composed with other nodes.

ct2034:
Hi @jlack1987 

Thanks for bringing this up. I agree that this would be useful. Especially when paired with the node that is diagnosed, this would save some overhead.

Pull requests are always welcome 🤗

PR #435, opened 2025-02-02
Make diagnostic aggregator composable

nnarain:
Heya.

This PR makes the `diagnostic_aggregator` node composable (see #321 )

Summary of changes:

- Have the `Aggregator` class implement `get_node_base_interface` so it can be registered as a component
- Register the aggregator as a `rclcpp_component`
- analyzer_group.cpp is moved to the analyzers library
  - The plugins xml file incorrectly lists the `AnalyzerGroup` plugin as existing the analyzers library. When composing the nodes I found that `AnalyzerGroup` would not load. I assume it has worked up to this point because everything was linked into the same executable.
  - Consequently this means the top level analyzer group is loaded as a plugin and therefore requires a class loader
- Added compose example launch file.

### DIFF
--- a/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.hpp
+++ b/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.hpp
@@ -134,6 +134,7 @@ public:
 
   DIAGNOSTIC_AGGREGATOR_PUBLIC
   rclcpp::node_interfaces::NodeBaseInterface::SharedPtr get_node_base_interface() const;
+
 private:
   rclcpp::Node::SharedPtr n_;
 
--- a/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.hpp
+++ b/diagnostic_aggregator/include/diagnostic_aggregator/aggregator.hpp
@@ -47,10 +47,11 @@
 #include <vector>
 
 #include "diagnostic_aggregator/analyzer.hpp"
-#include "diagnostic_aggregator/analyzer_group.hpp"
 #include "diagnostic_aggregator/other_analyzer.hpp"
 #include "diagnostic_aggregator/status_item.hpp"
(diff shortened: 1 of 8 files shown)


frozenreboot:
Hi @nnarain,

Is this PR still being worked on?
If you are busy, I’d be happy to pick it up, resolve the merge conflicts, and submit a new PR referencing your work.
Please let me know!

nnarain:
hey @frozenreboot 

Ya I should be able to get back to this. If I don't find the time this week I'll give you a ping.

Hey @ct2034 any initial feedback on this PR?

diagnostic_aggregator/include/diagnostic_aggregator/aggregator.hpp
```diff
@@ -102,7 +102,7 @@ other_as_errors: false
  * any analyzer is not properly specified, or returns false on initialization,
  * the aggregator will report the error and publish it in the aggregated output.
  */
-class Aggregator
+class Aggregator : public rclcpp::Node
```
Timple: This is not required for composable nodes. 
Franky: I would advise against it. It's hard if you want this class as a member of another class.

This pr would be much smaller if you kept the node member and the logger member.

nnarain: fair point on smaller PR

ct2034 (changes requested):
Thanks. This looks very good.
Could I please ask you to provide a small launch test. It could be based in the nice example `diagnostic_aggregator/example/compose_example.launch.py.in`.
And please also add a small section on this to the diagnostic_aggregator/README.md

nnarain:
Will do

nnarain:
There are two failing tests. But it looks like they are failing on the `ros2` branch as well.

PR #443, opened 2025-02-28
Fix reported hd_monitor available value

christophfroehlich:
I was wondering why hd_monitor reported 0.34%, but it was just not converted properly.
image

I propose to switch the (up to now) undocumented parameters to use percentage, otherwise it should be renamed to `ratio` instead of `percentage`. 

What do you think?


ct2034 (approved):
Thanks for the proposal. I think it works better this way

ct2034:
Can you please fix https://github.com/ros/diagnostics/blob/6e54e7fb2aa7f7dca32e1b1d23551a4335815a62/diagnostic_common_diagnostics/test/systemtest/test_hd_monitor_launchtest.py#L62

diagnostic_common_diagnostics/diagnostic_common_diagnostics/hd_monitor.py
```diff
@@ -82,8 +82,8 @@ def __init__(self):
         super().__init__(f'hd_monitor_{cleaned_hostname}')
 
         self._path = '~'
-        self._free_percent_low = 0.05
-        self._free_percent_crit = 0.01
+        self._free_percent_low = 5
+        self._free_percent_crit = 1
 
         self.add_on_set_parameters_callback(self.callback_config)
         self.declare_parameter('path', self._path)
```
ct2034: Could you please add explicit types here? Then the test would also work better

ct2034 (changes requested):

diagnostic_common_diagnostics/diagnostic_common_diagnostics/hd_monitor.py
christophfroehlich: you mean integer types for the parameters?

ct2034: What do you think? In practice an integer percentage should be enough, right? On the other hand it seems artificially restricting. But in your design you assumed integers, right?

christophfroehlich: Yes you are right, I changed it now to explicit as integer.
Side-effect: If someone had used this parameter before with the fraction/ratio, this now raises an error. (which is good)

christophfroehlich:
> Can you please fix
> 
sorry that I haven't run the tests before, should be good now.

ct2034 (approved):

ct2034:
Thanks for your contribution :) @christophfroehlich

christophfroehlich:
ament_flake8 should be happy now, too
```
$ ament_flake8 $(colcon list --packages-select diagnostic_aggregator diagnostic_common_diagnostics diagnostic_remote_logging diagnostic_updater self_test -p)

25 files checked
No problems found
```

christophfroehlich:
is there a chance to backport this to humble?

ct2034:
> is there a chance to backport this to humble?

Yes. But first we need to see what is wrong with that ci ..

ct2034:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|ros2-humble|[](https://github.com/ros/diagnostics/pull/454)|
|✅|ros2-kilted|[](https://github.com/ros/diagnostics/pull/455)|
|✅|ros2-jazzy|[](https://github.com/ros/diagnostics/pull/456)|

### Questions ?

PR #466, opened 2025-04-14
fix: publish diagnostic_agg using a timer, not wall_timer

MrBlenny:
Similar to https://github.com/ros/diagnostics/pull/458

diagnostics_agg should be published using a `timer` rather than `wall_timer` so it respects `use_sim_time`


ct2034 (approved):

redvinaa:
This doesn't solve the problem, because in the aggregator all status items create their own default clock.
https://github.com/ros/diagnostics/blob/61211e6cd8d88034ce3ee620eb67757b082d989f/diagnostic_aggregator/src/status_item.cpp#L49

ct2034:
@MrBlenny can you please open an issue, describe the problem you want to fix and how to reproduce it.

ct2034 (changes requested):
see comments above

PR #504, opened 2025-05-19
Update CMake requirements

mosfet80:
Cmake 3.5 is deprecaded.
Need to update requirements

Ros humble (oldest supported ros2) require cmake 3.20 and Cxx17.


ct2034:
#510

PR #506, opened 2025-05-23
Feature/report stale if one stale

Timple:
Going from:
> Report stale as errors unless all stale

to:
> If one STALE and no ERROR, report STALE

As a bonus, the message that caused the degradation of the diagnostics is added to the toplevel state. This is useful for reporting mechanisms such as error logs.

### DIFF
--- a/diagnostic_aggregator/src/aggregator.cpp
+++ b/diagnostic_aggregator/src/aggregator.cpp
@@ -217,6 +217,8 @@ void Aggregator::publishData()
   diag_toplevel_state.level = DiagnosticStatus::STALE;
   int max_level = -1;
   int min_level = 255;
+  int non_ok_status_depth = 0;
+  std::shared_ptr<DiagnosticStatus> msg_to_report;
 
   std::vector<std::shared_ptr<DiagnosticStatus>> processed;
   {
@@ -225,26 +227,56 @@ void Aggregator::publishData()
   }
   for (const auto & msg : processed) {
     diag_array.status.push_back(*msg);
+    const auto depth = std::count(msg->name.begin(), msg->name.end(), '/');
 
     if (msg->level > max_level) {
       max_level = msg->level;
+      non_ok_status_depth = depth;
+      msg_to_report = msg;
+    }
+    if (msg->level == max_level && depth > non_ok_status_depth) {
+      // On non okay diagnostics also copy the deepest message to toplevel state
+      non_ok_status_depth = depth;
+      msg_to_report = msg;
     }
     if (msg->level < min_level) {
       min_level = msg->level;
     }
+  }
+  // When a non-ok item was found, copy the complete status message once
+  if (max_level > DiagnosticStatus::OK) {
+    diag_toplevel_state.name = msg_to_report->name;
+    diag_toplevel_state.message = msg_to_report->message;
+    diag_toplevel_state.hardware_id = msg_to_report->hardware_id;
+    diag_toplevel_state.values = msg_to_report->values;
   }
 
   std::vector<std::shared_ptr<DiagnosticStatus>> processed_other =
     other_analyzer_->report();
   for (const auto & msg : processed_other) {
     diag_array.status.push_back(*msg);
+    const auto depth = std::count(msg->name.begin(), msg->name.end(), '/');
 
     if (msg->level > max_level) {
       max_level = msg->level;
+      non_ok_status_depth = depth;
+      msg_to_report = msg;
+    }
+    if (msg->level == max_level && depth > non_ok_status_depth) {
+      // On non okay diagnostics also copy the deepest message to toplevel state
+      non_ok_status_depth = depth;
+      msg_to_report = msg;
     }
     if (msg->level < min_level) {
       min_level = msg->level;
     }
+  }
+  // When a non-ok item was found, copy the complete status message once
+  if (max_level > DiagnosticStatus::OK) {
+    diag_toplevel_state.name = msg_to_report->name;
+    diag_toplevel_state.message = msg_to_report->message;
+    diag_toplevel_state.hardware_id = msg_to_report->hardware_id;
+    diag_toplevel_state.values = msg_to_report->values;
   }
 
   diag_array.header.stamp = clock_->now();
--- a/diagnostic_aggregator/src/aggregator.cpp
+++ b/diagnostic_aggregator/src/aggregator.cpp
@@ -216,7 +216,7 @@ void Aggregator::publishData()
   diag_toplevel_state.name = "toplevel_state";
   diag_toplevel_state.level = DiagnosticStatus::STALE;
   int max_level = -1;
-  int8_t max_level_without_stale = 0;
+  uint8_t max_level_without_stale = 0;
   int non_ok_status_depth = 0;
   std::shared_ptr<DiagnosticStatus> msg_to_report;
 
@@ -246,15 +246,7 @@ void Aggregator::publishData()
       max_level_without_stale = msg->level;
     }
   }
-  // When a non-ok item was found, copy the complete status message once
-  if (max_level > DiagnosticStatus::OK) {
-    diag_toplevel_state.name = msg_to_report->name;
-    diag_toplevel_state.message = msg_to_report->message;
-    diag_toplevel_state.hardware_id = msg_to_report->hardware_id;
-    diag_toplevel_state.values = msg_to_report->values;
-  }
 
-  non_ok_status_depth = 0;
   std::vector<std::shared_ptr<DiagnosticStatus>> processed_other =
     other_analyzer_->report();
   for (const auto & msg : processed_other) {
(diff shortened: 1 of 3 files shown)


Timple:
Rebased as build failed on https://github.com/ros/diagnostics/pull/507

Timple:
Any objections here?

We think it makes more sense to report STALE as summary. Because now you can have an ERROR without any actual ERROR states.

Timple:
If there are any reasons not to do this, we'd also like to hear. Or should this pass an MPC or something?

Timple:
Almost a year...

Can someone point us at someone who can shine a light if this is desired or undesired?

peci1:
I haven't gone through the logic of the implementation, but from a high-level point, I agree with this direction (I'm not a maintainer here, though).

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -225,39 +227,78 @@ void Aggregator::publishData()
...
 
   diag_array.header.stamp = clock_->now();
   agg_pub_->publish(diag_array);
 
-  diag_toplevel_state.level = max_level;
-  if (max_level < 0 ||
-    (max_level > DiagnosticStatus::ERROR && min_level <= DiagnosticStatus::ERROR))
+  if (
+    max_level == diagnostic_msgs::msg::DiagnosticStatus::STALE &&
+    max_level_without_stale < diagnostic_msgs::msg::DiagnosticStatus::ERROR)
```
mfaferek93: when no node has reported anything yet, `max_level` stays at `-1`, the new condition `max_level == STALE && ...` is false, and the `else` assigns `max_level_without_stale = 0 (OK)`, so a freshly-started aggregator with a dead pipeline publishes a green toplevel. The old code defended against this with `if (max_level < 0) level = ERROR;` - that guard is gone here.

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -225,39 +227,78 @@ void Aggregator::publishData()
...
+    if (
+      msg->level > max_level_without_stale &&
+      msg->level != diagnostic_msgs::msg::DiagnosticStatus::STALE)
+    {
+      max_level_without_stale = msg->level;
+    }
+  }
+  // When a non-ok item was found, copy the complete status message once
+  if (max_level > DiagnosticStatus::OK) {
+    diag_toplevel_state.name = msg_to_report->name;
```
mfaferek93: IMHO this silently breaks `diagnostic_remote_logging`. Its influxdb_connector.cpp https://github.com/ros/diagnostics/blob/ros2/diagnostic_remote_logging/src/influxdb_connector.cpp#L100 subscribes to `/diagnostics_toplevel_state` and uses `status.name` (via `splitName`) as the InfluxDB measurement name. Before this PR it was always `toplevel_state`. After this PR, on any degradation, the measurement renames to whatever analyzer is deepest (`Cell3`, `GPS`, etc.) and any dashboard keyed on `toplevel_state` loses its data.

mfaferek93: potential clean fix: keep `name = "toplevel_state"` stable and surface the offender via `message = msg_to_report->name + ": " + msg_to_report->message` (plus `hardware_id` and `values`). No consumer breaks.

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -225,39 +227,78 @@ void Aggregator::publishData()
...
-  if (max_level < 0 ||
-    (max_level > DiagnosticStatus::ERROR && min_level <= DiagnosticStatus::ERROR))
+  if (
+    max_level == diagnostic_msgs::msg::DiagnosticStatus::STALE &&
+    max_level_without_stale < diagnostic_msgs::msg::DiagnosticStatus::ERROR)
   {
-    // Top level is error if we got no diagnostic level or
-    // have stale items but not all are stale
-    diag_toplevel_state.level = DiagnosticStatus::ERROR;
+    diag_toplevel_state.level = diagnostic_msgs::msg::DiagnosticStatus::STALE;
```
mfaferek93: 1 corner case worth a sanity check. Concrete scenario: battery node reports `WARN` (low voltage), a USB camera drops at the same time and its node goes `STALE` after the watchdog timeout. Today the operator sees a single rollup on `/diagnostics_toplevel_state` and acts on it.

with this PR: `max_level = 3 (STALE)`, `max_level_without_stale = 1 (WARN)`, predicate `3 == STALE && 1 < ERROR` is true -> toplevel reports `STALE`. The battery WARN disappears. STALE reads like "camera reconnect, will resolve itself" and the live warning gets dismissed.

the old code reported `ERROR` here (`max_level > ERROR && min_level <= ERROR`). Also lossy, but at least it forced the operator to look deeper. Swapping it for `STALE` is a UX regression for that combination.

Potential fix:
```cpp
if (max_level_without_stale > DiagnosticStatus::OK) {
  level = max_level_without_stale;          // WARN/ERROR always beats STALE
} else if (max_level == DiagnosticStatus::STALE) {
  level = DiagnosticStatus::STALE;
} else if (max_level < 0) {
  level = DiagnosticStatus::STALE;          // empty input
} else {
  level = DiagnosticStatus::OK;
}
```

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -225,39 +227,78 @@ void Aggregator::publishData()
...
+  }
+  // When a non-ok item was found, copy the complete status message once
+  if (max_level > DiagnosticStatus::OK) {
+    diag_toplevel_state.name = msg_to_report->name;
+    diag_toplevel_state.message = msg_to_report->message;
+    diag_toplevel_state.hardware_id = msg_to_report->hardware_id;
+    diag_toplevel_state.values = msg_to_report->values;
   }
 
+  non_ok_status_depth = 0;
```
mfaferek93: The reset zeroes `non_ok_status_depth` but `msg_to_report` is kept from loop 1, so any non-OK message in `processed_other` at depth >= 1 will overwrite a deeper pick from `processed` (because `depth > 0` is trivially true after the reset), right?

diagnostic_aggregator/src/aggregator.cpp
```diff
@@ -216,7 +216,9 @@ void Aggregator::publishData()
   diag_toplevel_state.name = "toplevel_state";
   diag_toplevel_state.level = DiagnosticStatus::STALE;
   int max_level = -1;
-  int min_level = 255;
+  int8_t max_level_without_stale = 0;
```
mfaferek93: `max_level_without_stale` is `int8_t` here and `unsigned char` in `analyzer_group.cpp:295`. `DiagnosticStatus::level` is `int8` in the IDL, so `int8_t` is the canonical choice. Worth unifying :)

Timple:
Valid comments. Thank you for the review.

I managed to dispute none and to address them all :slightly_smiling_face:

diagnostic_aggregator/src/aggregator.cpp
Timple: Moved to this logic:
WARN/ERROR always beats STALE; if only STALE present, report STALE

diagnostic_aggregator/src/aggregator.cpp
Timple: This feature is backported from ROS1, there was no influxdb :smile: 
But name is reverted back to always report 'toplevel_state'.

mfaferek93 (approved):
LGTM!

ct2034:
Thanks for the review @mfaferek93 ! I agree with it and as far as I can follow it, the logic makes sense to me.
Is there a chance we could get an opinion from @asymingt ? (who developed this in https://github.com/ros/diagnostics/pull/315)

diagnostic_aggregator/test/test_discard_behavior.py
```diff
@@ -135,7 +135,7 @@
         foo_status=None,
         bar_discard=False,
         bar_status=None,
-        agg_expected=DiagnosticStatus.ERROR,  # <-- This is the case we are testing for.
+        agg_expected=DiagnosticStatus.STALE,  # <-- This is the case we are testing for.
         # if one of the children is *not* marked discard_stale := true and
         # there are no statuses, then the parent should roll up to ERROR.
```
ct2034:
```suggestion
        # there are no statuses, then the parent should roll up to STALE.
```

ct2034 (approved):
please fix this one comment

ct2034:
I made some more tests. ;-)
Could you please merge https://github.com/ros/diagnostics/tree/feature/report-stale-if-one-stale into your branch.
There is one test that is failing for me locally, and I would be interested in your opinion.

ct2034:
> I made some more tests. ;-) Could you please merge https://github.com/ros/diagnostics/tree/feature/report-stale-if-one-stale into your branch. There is one test that is failing for me locally, and I would be interested in your opinion.

Nvm, I think figured the test out. The case was wrong. But please still include them here @Timple.

ct2034:
Merged in #593

Timple:
I see I was too late. Thanks for merging.

Not sure about the backports though. It's a breaking change so not sure who would be affected in what way.

ct2034:
Nvm, it worked out.
I am always in favour of backports to keep the codebase consistent. And especially this case I would primarily treat as a bug, because the old behaviour seemed to not make sense to anyone.

PR #515, opened 2025-07-01
Fix node name and port in diagnostic_remote_logging README

christophfroehlich:
There are some issues in the README, it took me some time to set everything up (in the end it worked fine!)

- the node name is different, resulting in the parameter yaml not being used
- yaml syntax forbids keys without values
- the telegraf input listener needs a different port if the influxdb still uses the default 8086

### DIFF
--- a/diagnostic_remote_logging/README.md
+++ b/diagnostic_remote_logging/README.md
@@ -2,7 +2,7 @@ General information about this repository, including legal information and known
 
 # The diagnostic_remote_logging package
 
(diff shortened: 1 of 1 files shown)


ct2034:
@dwffls can you please look into this

dwffls:
All these changes are indeed correct. Made some mistakes with the creation of the README. 
So feel free to merge these!

@christophfroehlich Thanks for catching these, and happy to hear that the rest works fine for you!

dwffls:
@christophfroehlich I'm unsure why your node name differs from what's expected. According to [this line](https://github.com/ros/diagnostics/blob/7021f83f92251d08c075a31fc5d8ada1928ee247/diagnostic_remote_logging/CMakeLists.txt#L44)  in the CMakeLists.txt, the executable name, and therefore the node name, should be `influx`.

It's possible that you're naming it differently in your launch file. Could you check this?

The rest of your edits are excellent improvements to the README.md and I'd be happy to merge them, provided the node name is changed back to `influx`.

ct2034:
@dwffls The node name at runtime is defined by this line https://github.com/ros/diagnostics/blob/7021f83f92251d08c075a31fc5d8ada1928ee247/diagnostic_remote_logging/src/influxdb.cpp#L42C9-L42C17
But come to think of it, I don't like the name a lot anyway. The name `influxdb` suggests, it is the database itself. What do you think about `influxdb_connector` or something like that?

ct2034 (approved):
Thanks for the fix :)

PR #527, opened 2025-09-30
Fix windows build of diagnostic_remote_logging

christophfroehlich:
While working on a windows workflow for ros-controls, I encountered strange build errors from rclcpp. But it seems that the easier fix is to add `NOMINMAX` before including curl https://github.com/ros2/rclcpp/issues/2947#issuecomment-3339666233

Additionally, `unit_tests` did not build because no symbols were exported. There is no visibility control in the diagnostic_remote_logging, so I simply added `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`.


ct2034 (approved):
Thanks for your contribution

PR #545, opened 2025-12-02
Undefine ERROR macro when build on Windows

felixf4xu:
Undefine ERROR macro to build on Windows


ct2034 (changes requested):
1. Please open an issue and clearly describe the issue you are having. Including your OS version and all necessary steps
2. I will close this PR, because Error messages are an integral part of every ROS package and it builds in the CI as is.

PR #548, opened 2025-12-17
Add error percentage to cpu and ram monitor

sangteak601:
At the moment, it is not possible to configure `cpu_monitor` or `ram_monitor` to publish ERROR status. It would be good to have ERROR state. It would allow us to take actions before program exhibits unexpected behaviour.

Also, `cpu_monitor` was updated to compare threshold to average cpu usage rather than individual cpu usage. This is aligned with how `ram_monitor` works and makes more sense, as in most cases users are more interested in average usage than individual usage.

### DIFF
--- a/diagnostic_common_diagnostics/README.md
+++ b/diagnostic_common_diagnostics/README.md
@@ -13,6 +13,7 @@ It publishes the usage percentage in a diagnostic message.
 
 * Name of the node is "cpu_monitor_" + hostname.
 * Uses the following args:
+  * use_average: If true, the average CPU usage over all cores will be used to determine the status. If false, the maximum CPU usage among all cores will be used.
   * warning_percentage: If the CPU usage is > warning_percentage, a WARN status will be publised.
   * error_percentage: If the CPU usage is > error_percentage, an ERROR status will be published.
   * window: the maximum length of the used collections.deque for queuing CPU readings.
--- a/diagnostic_common_diagnostics/README.md
+++ b/diagnostic_common_diagnostics/README.md
@@ -14,6 +14,7 @@ It publishes the usage percentage in a diagnostic message.
 * Name of the node is "cpu_monitor_" + hostname.
 * Uses the following args:
   * warning_percentage: If the CPU usage is > warning_percentage, a WARN status will be publised.
+  * error_percentage: If the CPU usage is > error_percentage, a ERROR status will be published.
   * window: the maximum length of the used collections.deque for queuing CPU readings.
 
 ### Published Topics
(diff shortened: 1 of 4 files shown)


sangteak601:
@ct2034
It seems I can't add a reviewer. Would you be able to take a look at this?

diagnostic_common_diagnostics/diagnostic_common_diagnostics/cpu_monitor.py
```diff
@@ -71,15 +72,15 @@ def run(self, stat):
...
         for idx, cpu_percentage in enumerate(cpu_percentages):
             stat.add(f'CPU {idx} Load', f'{cpu_percentage:.2f}')
-            if cpu_percentage > self._warning_percentage:
-                warn = True
 
-        if warn:
+        if cpu_average > self._error_percentage:
+            stat.summary(DiagnosticStatus.ERROR,
+                         f'CPU Average exceeds {self._error_percentage} percent')
+        elif cpu_average > self._warning_percentage:
```
ct2034, line 81: This changes the behavior, because previously this warning was done per individual cpu. This is not bad per se. But the default behaviour should stay as it was previously

diagnostic_common_diagnostics/diagnostic_common_diagnostics/cpu_monitor.py
```diff
@@ -100,16 +101,20 @@ def main(args=None):
 
     # Declare and get parameters
     node.declare_parameter('warning_percentage', 90)
+    node.declare_parameter('error_percentage', 100)
```
ct2034, line 104: 100 is a weird default value. I suggest 95

diagnostic_common_diagnostics/README.md
```diff
@@ -14,6 +14,7 @@ It publishes the usage percentage in a diagnostic message.
 * Name of the node is "cpu_monitor_" + hostname.
 * Uses the following args:
   * warning_percentage: If the CPU usage is > warning_percentage, a WARN status will be publised.
+  * error_percentage: If the CPU usage is > error_percentage, an ERROR status will be published.
```
ct2034, line 17: please also make the this more precise in terms of average vs single cpu

diagnostic_common_diagnostics/diagnostic_common_diagnostics/cpu_monitor.py
```diff
@@ -51,10 +51,11 @@
 
 class CpuTask(DiagnosticTask):
 
-    def __init__(self, warning_percentage=90, window=1):
+    def __init__(self, warning_percentage=90, error_percentage=100, window=1):
```
ct2034, line 54:
```suggestion
    def __init__(self, warning_percentage, error_percentage, window):
```

ct2034: this way its closer to the ram implementation and there are not too many place where defaults are set.

ct2034 (changes requested):

ct2034:
@sangteak601 are you planning to work on this?

sangteak601:
> @sangteak601 are you planning to work on this?

All comments are addressed now I believe. Can you take another look?

ct2034 (changes requested):
I agree with all of the copilot comments

PR #551, opened 2026-03-05
Implement onParametersSet for handling only analyzers node parameters

Noel215:
The previous implementation is using a callback that is invoked for all the parameter events of all ROS 2 nodes and it filters by node name. It could lead to race conditions when setting parameters depending on the depth size of the callback.

With this new approach, the callback is only triggered when setting parameters for the analyzers node, avoiding possible race conditions when setting the parameters.


Noel215:
@ct2034 could you review/merge these changes?

ct2034 (approved):
Thanks for your contribution. This is a lot cleaner than before :)

